PR #85, opened 2025-05-08
chore: add all missing README files to repo

Parkreiner:
## Changes made
- Fleshed out main top-level README file
- Added formal docs for code of conduct and security (that just lead to the Coder Docs)
- Revamped contributing guide
- Added a few images to help support the new docs

## Notes
- Just because we're not supporting templates for the moment, I did deliberately limit the number of mentions to it.


bcpeinhardt (approved):

README.md
```diff
@@ -1,7 +1,50 @@
...
+
+- [The Cursor repo directory](https://github.com/coder/registry/tree/main/registry/coder/modules/cursor)
+- [The Cursor module page on the main website](https://registry.coder.com/modules/cursor)
+
+In both cases, the main README contains a Terraform snippet for integrating the module into your workspace. The snippet for Cursor looks like this:
+
+```tf
+module "cursor" {
+  count    = data.coder_workspace.me.start_count
+  source   = "registry.coder.com/modules/cursor/coder"
```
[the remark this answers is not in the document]
Parkreiner: Oh, good call. I ended up copying that snippet directly from the README, but now that you're calling it out, all the current README files still have the old path

@bcpeinhardt Just making sure: I just need to update all the README files to follow the pattern Atif suggested, right?

CONTRIBUTING.md
```diff
@@ -49,35 +49,87 @@ Once Go has been installed, verify the installation via:
 go version
 ```
 
-### Adding a new module/template (coming soon)
+## Namespaces
 
-Once Bun (and possibly Go) have been installed, clone this repository. From there, you can run this script to make it easier to start contributing a new module or template:
+All Coder resources are scoped to namespaces placed at the top level of the `/registry` directory. Any modules or templates must be placed inside a namespace to be accepted as a contribution. For example, all modules created by TheJonkler would be placed under `/registry/thejonkler/modules`, with a subdirectory for each individual module the user has published.
```
f0ssel, line 54: Let's use an example that isn't close to anything real just to be safe

CONTRIBUTING.md
```diff
@@ -133,27 +185,60 @@ The following criteria exists for one of two reasons: (1) content accessibility,
...
+
+- Frontmatter metadata with support for the following fields:
+
+  - `display_name` (required string) – The name to use when displaying your user profile in the Coder Registry site.
+  - `bio` (optional string) – A short description of who you are.
+  - `github` (optional string) – Your GitHub handle.
+  - `avatar_url` (optional string) – A relative/absolute URL pointing to your avatar for the Registry site. It is strongly recommended that you commit avatar images to this repo and reference them via a relative URL.
+  - `linkedin` (optional string) – A URL pointing to your LinkedIn page.
+  - `support_email` (optional string) – An email for users to reach you at if they need help with a published module/template.
+  - `status` (optional string union) – If defined, this must be one of `"community"`, `"partner"`, or `"official"`. `"community"` is treated as the default value if not specified, and should be used for the majority of external contributions. `"partner"` is for companies who have a formal business partnership with Coder. `"official"` should be used only by Coder and Coder satellite companies.
```
f0ssel, line 200: Instead of an implicit default, can we have an explicit set and fail if missing?

f0ssel: Do we need "and Coder satellite companies"?

README.md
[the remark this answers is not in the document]
bcpeinhardt: Yes. The old path will still work, because we've added a shim to avoid breaking peoples workspace builds, but yeah the namespace is now "coder" not "modules"

CONTRIBUTING.md
Parkreiner: Addressed both of these, and updated the script, too

Parkreiner:
Addressed all the feedback. I don't want to update the source URLs now, since we haven't pushed the latest changes to the production Registry site, but I went ahead and made a separate PR to handle that.

Issue #27, opened 2025-04-23
RFC(Coder-Server): Template Settings


djarbz:
Referring to the `settings` option on the module.
https://github.com/coder/modules/blob/4b4cebaf4dfbf200ff9cb2a0dc6ff9bbf3101ce8/code-server/run.sh#L22C1-L27C3

My understanding is that there are 3 levels of preference settings and highest takes precedence.

1. Workspace settings: `<GIT REPO>/.vscode/settings.json`
2. User settings: `$HOME/.local/share/code-server/User/settings.json`
3. Machine settings: `$HOME/.local/share/code-server/Machine/settings.json`

I would put `1` & `2` as user defined settings where a dev might need to configure a setting that should be shared by a team and personal settings for their own personal preference.
I would then think that a template owner would want to provide sensible defaults in `3` that can then be overridden by the user via option `1` & `2`.
Currently this module writes to `2` upon workspace creation, but never again, even if the template is updated.

My proposal is that, for backwards compatibility, a new `machine-settings` option is added to the module, much like the current `settings` option, except that it will always be overwritten by the module upon startup.
This will allow template owners to create sensible defaults and allow updating and changing these defaults in future template revisions.
This should also allow users and repos to override these settings via the `user` and `workspace` settings.

PR #88, opened 2025-05-09
feat(code-server): add machine settings option

djarbz:
Should resolve https://github.com/coder/registry/issues/27

The goal is to provide a method for template owners to configure default `Machine Settings` that can be overridden by developers via `User Settings` and repositories via `Workspace Settings`.
This option allows template owners to push new settings options with a template release that would not be ignored because the setting file already exists.
This also formats the `settings.json` file if `jq` is installed.

Eventually, I would imagine the current `settings` option will be depreciated in favor of this option.

### DIFF
--- a/registry/coder/modules/code-server/main.tf
+++ b/registry/coder/modules/code-server/main.tf
@@ -44,6 +44,12 @@ variable "settings" {
   default     = {}
 }
 
+variable "machine-settings" {
+  type        = any
+  description = "A map of template level machine settings to apply to code-server. This will be overwritten at each container start."
+  default     = {}
+}
+
 variable "folder" {
   type        = string
   description = "The folder to open in code-server."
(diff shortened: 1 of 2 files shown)


DevelopmentCats:
I'm going to test this out in a template, and then I will update.

DevelopmentCats:
I was taking another look at everything before testing, I noticed something about the settings.json formatting.  You are using jq when it's installed, which is great, but maybe we should check if jq is available first and install it automatically if it's missing? That way we can ensure consistent formatting for all users rather than having it depend on whether they happen to have jq installed already.

djarbz:
I'm hesitant to add a dependency just for that.
Updating the package list and installing takes time, I would rather a quick stand-up than to have to wait for jq to install.
Formatting is moreso a nice-to-have, I don't expect many to manually modify the file and rather edit from the settings menu.

DevelopmentCats:
Okay that works.

I'll test this when I get home in a bit and then I'll merge.

DevelopmentCats (approved):
Tested and its working fine.

DevelopmentCats:
Need to add Versions in README for code-server

djarbz:
@DevelopmentCats Do I need to update the versions? From what I had read in the contributors.md, that was handled internally. Unless I am mistaken?

PR #90, opened 2025-05-11
feat(vault-token): make supplying a vault token optional

moo-im-a-cow:
makes the vault token variable optional,
meaning that template creators can use the vault token module just to install vault cli, without it configuring a token environment variable.
the template creator can then log into vault with their own script,
or they can ask the user to manually run vault login command

potential future change: refactor the vault-jwt and the vault-github modules in this repo to call this module for installing vault, and their scripts only need to call the relevant login command, they dont need to download and install vault. makes maintaining easier if you have just one install script.

completely backwards compatible


matifali:
@bcpeinhardt, what is the best way to version this module? Should we move to the new namespaced format and start from 2.0.0? It can be 1.2.0, i.e, the next version as per the semver, but in a new format.

DevelopmentCats (approved):
Looks good to me

PR #89, opened 2025-05-09
Mes/new readme files 2

Parkreiner:

PR #93, opened 2025-05-12
chore: update module sources for community modules

matifali:
updates #89


registry/nataindata/README.md
```diff
@@ -3,5 +3,5 @@ display_name: Nataindata
 bio: Data engineer
 github: nataindata
 website: https://www.nataindata.com
-status: partner
+status: community
```
Parkreiner, line 6: Oh shoot, forgot about that. Thanks for catching that, Atif!

Parkreiner (commented):
Changes look good to me, but I feel like @bcpeinhardt should be the gatekeeper here

I swear we said that we were going to wait until after we pushed a meaningful change to a module to bump the version and switch the import path. But I guess there's a risk that if some of these modules go a while without updates (Airflow has had 7 total for the entire lifetime of the Registry), that increases the risk we forget

We'd also have to push a new tag for each individual module, right?

registry/nataindata/README.md
bcpeinhardt: @Parkreiner we just need to push tags for the community modules as they get properly new namespaces (whereas "coder" and "modules" are just being treated as the same namespace).
I agree from a programming perspective I finding bumping the version without changing the source of the module annoying but I suppose it's better to do this now as we're "launching" partner and community namespacing rather than have a drip of "x modules is moving to a new namespace" over the course of six months.
We used to bump all the versions together so folks are kind of used to jumping several patch versions anyway.

bcpeinhardt (approved):

PR #96, opened 2025-05-14
chore: migrate new_module script

f0ssel:
Closes https://github.com/coder/internal/issues/611

This scripts creates a new sample moduledir with required files
Run it like : ./scripts/new_module.sh my-namespace/my-module


scripts/new_module.sh
matifali, line 1: I think we now also need a sample `NAMESPACE` README.md

matifali (approved):
Thanks Garret

matifali:
@f0ssel could you also update CONTRIBUTING.md?
https://github.com/coder/registry/blob/main/CONTRIBUTING.md#coder-modules
Thanks,. We missed that in the PR.

PR #97, opened 2025-05-15
chore: remove support email

matifali:
The email is only for paying customers. Removing this will help us get random support requests.


DevelopmentCats (approved):
Sounds like a time-saver to me!

PR #106, opened 2025-05-16
fix(coder/modules/goose): update default values for Goose provider and model variables to empty strings

DevelopmentCats:
- Changed default values from null to empty strings for `experiment_goose_provider` and `experiment_goose_model` variables in main.tf to ensure compatibility and avoid null interpolation issues.


matifali (commented):
I am assuming you tested that this change fixes the issue.
Also can you please bump the patch version?

ericpaulsen (approved):

DevelopmentCats:
Yeah Ill bump the version now and I have tested it in my environment pulling from the branch.

But now i'm getting npm install issues all of a sudden so I'm going to look into that real quick

matifali (approved):

PR #107, opened 2025-05-16
refactor: use coder/slog + minor go style changes

cstyan:
Changes are broken down in to multiples commits to hopefully make reviewing easy. 1 commit for the slog change and then a commit per Go file for style changes.

Style changes are generally:
- try to use full sentences for all comments
- try to stick to 120 column lines (not strict) instead of 80
- try to one line as many `call function, check if err != nil` blocks as possible
- stick var and const definitions near the top of the file
- try to use `err` or `errs` for all return type names, previously used `problems` in some cases but `errs` in others
- some minor optimizations, like the line scanner declaring a new variable in each iteration of a loop
- `Todo` -> `TODO`, sometimes also useful to do `TODO (name):` to make it easier to find things a specific author meant to follow up on
- comments for types/functions should generally start with `// FunctionName/TypeName ...` though I'm now seeing places I didn't update that

In general there's very few tests for the Go code here, would we like more or is there some testing that spins up the entire registry to validate things? I didn't see any makefile.


cmd/readmevalidation/readmefiles.go
```diff
@@ -8,48 +8,73 @@ import (
...
+	// being parsed as YAML. This phase does not include YAML validation.
+	validationPhaseReadmeParsing = "README parsing"
 
-// readme represents a single README file within the repo (usually within the
-// top-level "/registry" directory).
+	// validationPhaseAssetCrossReference indicates when a README's frontmatter
+	// is having all its relative URLs be validated for whether they point to
+	// valid resources.
+	validationPhaseAssetCrossReference = "Cross-referencing relative asset URLs"
+)
```
Parkreiner, line 32: On the topic of using declarations to group related variables, I feel like I'd want two `const` declarations here – one for the phases, and one for everything else

Parkreiner: Two other things:
- I'm hesitant about referencing something above where it's defined if I can help it (in this case, defining a string in terms of `validationPhase` before the type is declared), even if it'll still work. I generally like being able to read the code top-to-bottom, even if that means that files don't follow a pattern of always defining constants first
- The phases were previously defined as ints via iota. I'm just now realizing: with the current setup, only the first phase is defined strictly as `validationPhase`, right? Everything else in the series has the type `untyped string`?

cmd/readmevalidation/contributors.go
```diff
@@ -245,11 +240,12 @@ func aggregateContributorReadmeFiles() ([]readme, error) {
 	allReadmeFiles := []readme{}
 	errs := []error{}
 	for _, e := range dirEntries {
-		dirPath := path.Join(rootRegistryPath, e.Name())
 		if !e.IsDir() {
 			continue
 		}
 
+		dirPath := path.Join(rootRegistryPath, e.Name())
```
Parkreiner, line 247: Oh, good call

cmd/readmevalidation/contributors.go
```diff
@@ -318,19 +310,18 @@ func validateAllContributorFiles() error {
 		return err
 	}
 
-	log.Printf("Processing %d README files\n", len(allReadmeFiles))
+	logger.Info(context.Background(), "Processing README files", "num_files", len(allReadmeFiles))
```
Parkreiner, line 313: I'm still new to structured logging. Is there any special behavior/benefit you get if you use the same key multiple times? I guess I'm just wondering how much of a concern it is to make sure you're using the same keys each time you describe the same resource, particularly for a function call that takes a variadic slice of empty interfaces (so basically zero type-safety)

cmd/readmevalidation/readmefiles.go
```diff
@@ -8,48 +8,73 @@ import (
...
+
+var (
+	supportedAvatarFileFormats = []string{".png", ".jpeg", ".jpg", ".gif", ".svg"}
+	// TODO: an example of what this regex matches would be useful, I think it's just
+	readmeHeaderRe = regexp.MustCompile("^(#{1,})(\\s*)")
```
Parkreiner, line 37: I think the comment didn't get filled out, but it's meant to match lines that start like this (the single space is required):
- `# `
- `## `
- `### `
- `#### `
- `##### `
- `###### `

cmd/readmevalidation/readmefiles.go
```diff
@@ -8,48 +8,73 @@ import (
...
-// validate whether the structure of the frontmatter is valid (i.e., that it's
+// separateFrontmatter attempts to separate a README file's frontmatter content from the main README body, returning
+// both values in that order. It does not validate whether the structure of the frontmatter is valid (i.e., that it's
 // structured as YAML).
 func separateFrontmatter(readmeText string) (string, string, error) {
 	if readmeText == "" {
 		return "", "", errors.New("README is empty")
 	}
 
-	const fence = "---"
```
Parkreiner, line 31: Is there any harm to keeping `fence` confined to this function? It's only relevant for this one function right now, and I was under the impression that if you had a const declaration defined in a function, it would only be created once at build time, but only be accessible to the function

cmd/readmevalidation/readmefiles.go
```diff
@@ -66,20 +91,17 @@ func separateFrontmatter(readmeText string) (string, string, error) {
...
 	trimmed := strings.TrimSpace(body)
 
 	if trimmed == "" {
 		return []error{errors.New("README body is empty")}
 	}
 
-	// If the very first line of the README, there's a risk that the rest of the
-	// validation logic will break, since we don't have many guarantees about
-	// how the README is actually structured
+	// If the very first line of the README, there's a risk that the rest of the validation logic will break, since we
```
Parkreiner, line 103: Oh, I forgot some words. This was supposed to be "If the very first line of the README doesn't start with an ATX-style H1 header"

Parkreiner (commented):
Just commenting for now, since it sounded like there might be some more changes you want to make, but I'm okay with making these changes

And to be clear, when I'm asking a question, that's not me trying to be defensive – I'm just trying to understand how big the gap between my TypeScript way of doing things  is with how a Gopher usually does stuff

cmd/readmevalidation/coderresources.go
```diff
@@ -124,15 +122,15 @@ func validateCoderResourceReadmeBody(body string) []error {
 	lineNum := 0
 	isInsideCodeBlock := false
 	isInsideTerraform := false
+	nextLine := ""
 
 	lineScanner := bufio.NewScanner(strings.NewReader(trimmed))
 	for lineScanner.Scan() {
 		lineNum++
-		nextLine := lineScanner.Text()
+		nextLine = lineScanner.Text()
```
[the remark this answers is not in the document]
cstyan: Yeah this is just an optimization to reduce memory allocations. Very minor in this case since I doubt this loop has a lot of iterations, but without this a new string for `nextLine` is allocated for each iteration of the loop. 

The Go compiler already does an optimization itself for `for thing := range anotherThing` to do the same optimization, assigning to the same var for each iteration rather than allocating a new one every time.

cmd/readmevalidation/contributors.go
cstyan: It's not the end of the world if you don't use the same key, but it does make searching for logs in some kind of log aggregation system much easier.

For example, a system I used to work on referred to the same internal `tenant` type within the system as variations of `user`, `tenant`, `id`, etc. Remembering which key was used on which logged lines complicated searches when I knew within I needed to see info for `tenant="1234"` but on some lines the logging was `user="1234"`.

Again this is likely less important in the case of the registry but still a good practice.

cmd/readmevalidation/readmefiles.go
cstyan: > On the topic of using declarations to group related variables, I feel like I'd want two const declarations here – one for the phases, and one for everything else

Can you elaborate as to why? It's idiomatic, but not required, to group together const/var definitions like this. We could separate the const definitions, or simply separate them within the same parens block with a comment. 

> I'm hesitant about referencing something above where it's defined if I can help it (in this case, defining a string in terms of validationPhase before the type is declared), even if it'll still work. I generally like being able to read the code top-to-bottom, even if that means that files don't follow a pattern of always defining constants first

I agree with you, for sure. I missed that `validationPhaseFileStructureValidation` was a `validationPhase`. I'll see if there's another refactor we can make here to keep the top to bottom organization in place.

> The phases were previously defined as ints via iota. I'm just now realizing: with the current setup, only the first phase is defined strictly as validationPhase, right? Everything else in the series has the type untyped string?

Well, string type no untyped, but yes. How important is it that the `validationPhase` exist? Could we use `iota` again or have these all just be strings?

cmd/readmevalidation/readmefiles.go
cstyan: Good point, I will move it back.

cmd/readmevalidation/repostructure.go
```diff
@@ -12,40 +12,37 @@ import (
 var supportedUserNameSpaceDirectories = append(supportedResourceTypes[:], ".icons", ".images")
 
 func validateCoderResourceSubdirectory(dirPath string) []error {
-	errs := []error{}
+	var (
+		err    error
+		subDir os.FileInfo
+	)
```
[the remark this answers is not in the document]
cstyan: Personal preference in some cases. The convention is _either_ if the variables are logically related to each other, or to help with readability such as when there's multiple variables declared near eachother and you want to avoid repeating the `var` keyword.

In this case, I wanted both to get their respective default values and allow for turning the previous lines 17-18 into one line.

The other option was to do:
```
var subDir os.FileInfo
var err Error
```

cmd/readmevalidation/repostructure.go
```diff
@@ -12,40 +12,37 @@ import (
...
 	}
 
 	files, err := os.ReadDir(dirPath)
 	if err != nil {
-		errs = append(errs, addFilePathToError(dirPath, err))
-		return errs
+		return []error{addFilePathToError(dirPath, err)}
 	}
+
+	errs := []error{}
```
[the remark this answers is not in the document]
cstyan: > Question I've had for a bit: does it matter whether errs is defined with an allocation or as a nil slice, since we're not serializing it as JSON?

Do you mean empty slice, as opposed to nil slice? Rather than doing `var slice := make(...)`?

Usually I prefer doing `make(...)` with some specified length/capacity since that allows for either starting out with a slice of the size you need, or at least of some reasonable size. Every call to `append` when the underlying memory no longer has remaining capacity for what you're trying to append results in reallocation of the slice with 2x the current capacity.

When we don't know what the final length might be and it is possible that it could be 0, using `[]error{}` ensures we don't allocate any space for the item storage portion of the slice.

Even if we did `errs := make([]error, 0, 10)` the underlying item storage would still be allocated for 10 items.

In general it's best to avoid nil slices for return values, though they can be used for function parameters/optional values.

One important point to note is that under the hood you can append to a nil slice, it will be treated as a 0 length empty slice on the first append.

cmd/readmevalidation/repostructure.go
```diff
@@ -54,27 +51,28 @@ func validateCoderResourceSubdirectory(dirPath string) []error {
...
 }
 
 func validateRegistryDirectory() []error {
-	userDirs, err := os.ReadDir(rootRegistryPath)
-	if err != nil {
+	var (
+		userDirs []os.DirEntry
+		err      error
+	)
+	if userDirs, err = os.ReadDir(rootRegistryPath); err != nil {
```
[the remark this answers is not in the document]
cstyan: Just to allow for the assignment and err check in one line, while still being able to use the value of `userDirs` later on in the function. I may have gone a bit overboard with this type of change in this PR :joy: 

> I feel like we got a lot out of that : before, especially the type inference it gives

True, though it's safe in this case since we're calling a function from the stdlib. The function signature for `os.ReadDir` will not change without a major version change in Go. If changes were required before a Go 2.0, a new function would be introduced rather than `os.ReadDir` being modified in a way that changed it's return types.

bcpeinhardt:
@cstyan 
1. To your point on the tests, I think this was quite quickly written validation code meant to duplicate validation done in the coder/registry-server repo, so the larger testing story of the registry is located there. That said, we should have this code live in one place where it is well tested and import it where we need it.
2. I wonder how many of these we can lint in our local dev process? I think it'd be awesome to come to a team preference on stuff like "prefer handling `err != nil` checks inline when possible" and then lint for those things.

Parkreiner:
@bcpeinhardt Yeah, the thinking at the time was that because we had so many real-world README files already, we could use them as an implicit test case. The only reason why there's that one test case now is because it made development easier when I was writing the functionality

cmd/readmevalidation/coderresources.go
```diff
@@ -338,17 +335,16 @@ func validateAllCoderResourceFilesOfType(resourceType string) error {
...
 	resources, err := parseCoderResourceReadmeFiles(resourceType, allReadmeFiles)
 	if err != nil {
 		return err
 	}
-	log.Printf("Processed %d README files as valid Coder resources with type %q", len(resources), resourceType)
+	logger.Info(context.Background(), "Processed README files as valid Coder resources", "num_files", len(resources), "type", resourceType)
 
-	err = validateCoderResourceRelativeUrls(resources)
-	if err != nil {
+	if err = validateCoderResourceRelativeUrls(resources); err != nil {
```
cstyan, line 345: URLs

cmd/readmevalidation/contributors.go
cstyan: the initial creation could actually outside the loop and this would just be assignment

cmd/readmevalidation/contributors.go
```diff
@@ -318,19 +310,18 @@ func validateAllContributorFiles() error {
...
 	contributors, err := parseContributorFiles(allReadmeFiles)
 	if err != nil {
 		return err
 	}
-	log.Printf("Processed %d README files as valid contributor profiles", len(contributors))
+	logger.Info(context.Background(), "Processed README files as valid contributor profiles", "num_contributors", len(contributors))
 
-	err = validateContributorRelativeUrls(contributors)
-	if err != nil {
+	if err = validateContributorRelativeUrls(contributors); err != nil {
```
cstyan, line 320: URLs

cmd/readmevalidation/main.go
```diff
@@ -6,20 +6,23 @@
...
-	log.Println("Starting README validation")
+	logger.Info(context.Background(), "Starting README validation")
 
-	// If there are fundamental problems with how the repo is structured, we
-	// can't make any guarantees that any further validations will be relevant
-	// or accurate
+	// If there are fundamental problems with how the repo is structured, we can't make any guarantees that any further
+	// validations will be relevant or accurate.
 	repoErr := validateRepoStructure()
 	if repoErr != nil {
```
cstyan, line 24: one line?

cmd/readmevalidation/readmefiles.go
```diff
@@ -109,8 +132,7 @@ func validateReadmeBody(body string) []error {
 			continue
 		}
 
-		spaceAfterHeader := headerGroups[2]
-		if spaceAfterHeader == "" {
+		if spaceAfterHeader := headerGroups[2]; spaceAfterHeader == "" {
```
cstyan, line 135: don't need the var

cmd/readmevalidation/coderresources.go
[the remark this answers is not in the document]
Parkreiner: Is that actually true nowadays? I know it used to be, but they changed the `range` behavior to isolate variables for each declaration in Go 1.22:
https://go.dev/blog/loopvar-preview

Is Go doing escape analysis for the variable to see if it needs to be kept around for closures, and only reusing the declaration if it knows that the variable doesn't need to be long-lived? JS enforces scoped behavior for every loop that uses let and const, but I imagine it can't do those optimizations natively because it's a JIT language

cmd/readmevalidation/repostructure.go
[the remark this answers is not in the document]
Parkreiner: To be clear, I mean the difference between:
```go
var errs []error
```
Which is always created as `nil` and never involves upfront allocations. If I understand right, the allocations only happen behind the scenes if you pass the value into an `append` call, and if you never call `append`, the slice lives exclusively on the stack

and
```go
errs := []error{}
```
Which always causes an allocation, just with a backing array of length 0

cmd/readmevalidation/readmefiles.go
Parkreiner: > Can you elaborate as to why?
I value the ease of reading code over the ease of writing it, always. I'll concede if it's more idiomatic, but while you can group a bunch of declarations together, but then there's zero semantic information being conveyed when they're all under one group – there's no meaningful hierarchy, aside from whether they can be reassigned. With more logical groups (and the indentation that comes with it), it's easier to skim over the code and find elements that are meant to be related. That might be more of a take from someone who isn't leveraging IDE features enough, though

> Well, string type no untyped, but yes
To be clear, I'm talking about this behavior:
```go
// Go infers this as the base type string at a minimum, but it appends the `untyped`
// modifier to it. You can see this by hovering over the definition in an IDE
const x = "blah"

// These two types are represented as strings under the hood, but they're very obviously
// different. In the real world, you'd probably want to define parse functions alongside them
type PhoneNumber string
type FavoriteColor string

func Test() {
	var huh string
	var num PhoneNumber
	var color FavoriteColor
	
        // Allowed
	num = x

        // Allowed
	color = x

        // Now allowed
	num = huh

        // Now allowed
	color = num
}
```

> How important is it that the validationPhase exist? Could we use iota again or have these all just be strings?
I feel like that might be more of a @f0ssel question

cmd/readmevalidation/coderresources.go
Parkreiner: I'm still trying to wrap my head around the benefits of this pattern now that we don't have a `:=` assignment. I'm generally with @f0ssel in that I'm not a big fan of the pattern – except when it gives you the ability to enforce more aggressive block-based scoping

With `err` defined outside the block, though, this feels like shorthand for the sake of shorthand. It's no longer giving you any new runtime behavior, and it doesn't feel like a meaningful refactor to me

Please tell me if I'm missing something, though

Parkreiner (approved):
I'm approving because I trust you, and I want to be clear: I hope my questions/skepticism  don't come across as personal attacks towards you. I trust you – I think I might just be going through some culture shock as I figure out how to be a gopher myself

I do think it might be good to get some extra input from @f0ssel on some of the changes you were asking about before merging, though. He was the one who originally reviewed the code as I was submitting it

cmd/readmevalidation/coderresources.go
cstyan: Not sure I understand, the existing code having two lines for assignment to an existing var then a second for the if condition. This is the exact case  where it's idiomatic to reuse the existing declaration of `err` via `err =` as opposed `err :=` and one line the assignment + error check since this instance of the err particular value is only used within the if condition for the return value.

There may be other cases where I went over board with shorthands, but in this specific case there's no benefit to an additional line or new `err := ...`. Linters will explicitly complain about the latter as it shadows the earlier declaration of the same variable.

Did @f0ssel leave some feedback about this pattern in the past somewhere?

cmd/readmevalidation/repostructure.go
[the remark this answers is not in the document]
cstyan: Yep you're correct. The slice header exists on the stack, and with the `var errs []error` version we have just the slice header (24 bytes) and the pointer to the backing array is nil since we don't have one yet.

In the case of `errs := []error{}` the backing array is initialized on the heap, but I believe that it's using at most 8 bytes until we do some amount of appends to the slice.

IMO, and happy to discuss whether we want to use this pattern or something else, is that it's always better to _not_ have the potential for a nil slice to avoid the potential for attempted indexing into a nil slice. Whether that be in our actual code just within a test.

cstyan:
> I'm approving because I trust you, and I want to be clear: I hope my questions/skepticism don't come across as personal attacks towards you. I trust you – I think I might just be going through some culture shock as I figure out how to be a gopher myself
> 
> I do think it might be good to get some extra input from @f0ssel on some of the changes you were asking about before merging, though. He was the one who originally reviewed the code as I was submitting it

Not at all, I appreciate the discussion and being challenged on some of this. It's a good reminder that a) some of what Go does feels overly prescriptive or pedantic from the outside without giving clear reasons why, and it's a good reminder for me to reevaluate whether my own understanding of the why is still valid or it's just semantic convention because the style guide says so.

There may also be some style changes I've applied here that are in addition to the Go or Google conventions, IIRC we do have a Coder Go style guide somewhere, I'll have another read through of that today.

cmd/readmevalidation/readmefiles.go
cstyan: > I value the ease of reading code over the ease of writing it, always. I'll concede if it's more idiomatic for Go, but while you can group a bunch of declarations together, but then there's zero semantic information being conveyed when they're all under one group – there's no meaningful hierarchy, aside from whether they can be reassigned. With more logical groups (and the indentation that comes with it), it's easier to skim over the code and find elements that are meant to be related. That might be more of a take from someone who isn't leveraging IDE features enough, though

Yep, I agree. I value the same. To me there's just as much semantic meaning conveyed by grouping them together within one const block but separating them by comments as there would be by having separate const groups. And in this case these are all package level constants.

If these specific constants were defined within a function and not used (or redeclared) anywhere else then I agree we should have left them as is. Since both the type and constants were package it (to me) reads more easily to keep them at the top of the file and in one block.

I do think we can do a bit more of a refactor here as currently it's inconsistent as to whether each is an instance of a `validationPhase` or just a concrete string, even though all are named with the prefix `validationPhase...`.

> To be clear, I'm talking about this behavior:

Thanks for clarifying. Yep, that's definitely true that typing the string like this is useful. At the very least, it forces someone to type cast if they insist on doing something that might be slightly unsafe/cause unexpected behaviour.

cmd/readmevalidation/coderresources.go
Parkreiner: I don't want to speak for f0ssel, but the way I see it is that generally, code is the most clear and easy to refactor when it does as little as possible per line

Assigning to a variable and immediately evaluating it on one line breaks from that, but I'm generally okay with it because the variable in question is usually also initialized on the line, and Go forces the variable to be scoped to the block. Without an initialization, though, it feels like we're merging lines together simply because Go lets us, and not strictly because it gives us any benefits around behavior/access or readability

Parkreiner:
To be honest, I didn't even realize that we had a backend contributing guide, since it's not in the Docs site. I thought we only had a frontend contributing guide

cmd/readmevalidation/readmefiles.go
Parkreiner: > I do think we can do a bit more of a refactor here as currently it's inconsistent as to whether each is an instance of a validationPhase or just a concrete string, even though all are named with the prefix validationPhase....

Yeah, that's fair. I think I was going for something similar to the setup for the status codes in the `http` package (`StatusOK`, `StatusConflict`, etc.), but even those are just represented as plain ints. I do feel like the autocomplete experience is handy, though, since you just have to know the prefix to get the IDE to display the full list

I'm down for whatever's more idiomatic, though

cstyan:
> To be honest, I didn't even realize that we had a backend contributing guide, since it's not in the Docs site. I thought we only had a frontend contributing guide

I may be misremembering, I can't find that document now.

cmd/readmevalidation/readmefiles.go
cstyan: > I do feel like the autocomplete experience is handy, though, since you just have to know the prefix to get the IDE to display the full list

This is a good point 👍 The main thing I wanted to resolve was the mismatching types.

What about the following for names:
```
// validationPhaseStructure indicates when the entire Registry
// directory is being verified for having all files be placed in the file
// system as expected.
validationPhaseStructure validationPhase = "File structure validation"

// validationPhaseFile indicates when README files are being read from the file system.
validationPhaseFile validationPhase = "Filesystem reading"

// validationPhaseReadme indicates when a README's frontmatter is
// being parsed as YAML. This phase does not include YAML validation.
validationPhaseReadme validationPhase = "README parsing"

// validationPhaseCrossReference indicates when a README's frontmatter
// is having all its relative URLs be validated for whether they point to
// valid resources.
validationPhaseCrossReference validationPhase
```

cmd/readmevalidation/coderresources.go
Parkreiner: Honestly, I'm okay with whatever feels more idiomatic. When I had my 1:1 with Garrett last week, I asked more about where my own misunderstandings were coming from, and I think I have a better sense for what Go engineers typically think about now

I'm still not a fan, but if it's a pattern that most Go devs do, I don't want to keep bike-shedding on this

cmd/readmevalidation/coderresources.go
```diff
@@ -105,16 +108,11 @@ func validateCoderResourceTags(tags []string) error {
...
-// it's structured. Just validating whether it *can* be parsed as Terraform
-// would be a big improvement.
-var terraformVersionRe = regexp.MustCompile("^\\s*\\bversion\\s+=")
-
 func validateCoderResourceReadmeBody(body string) []error {
-	trimmed := strings.TrimSpace(body)
 	var errs []error
+
+	trimmed := strings.TrimSpace(body)
+	// TODO: this may cause unexpected behaviour since the errors slice may have a 0 length. Add a test.
```
cstyan: can remove this comment, under the hood the go runtime will handle allocating a slice of length 1 if the length at the time of an append is 0

PR #108, opened 2025-05-18
feat(vault-token): add optional vault enterprise namespace variable

moo-im-a-cow:
Added an optional envvar to vault-token module to handle communicating with a non default vault namespace.

in vault enterprise, you can run multiple secure isolated vault environments from the one vault server. 
each namespace has it's own authentication methods and secrets engines. 
vault uses the VAULT_NAMESPACE envvar to determine the namespace to use.
no value, or either `root` or `/` will use the root (default) namespace, any other value will use a different namespace

in vault community edition, the only supported namespace is "root", no other namespaces can be used.

in HCP vault dedicated (the saas hosted version), you cant access vault without a namespace set

this defaults to not setting the env var, so is backwards compatible, and works with vault CE

### DIFF
--- a/registry/coder/modules/vault-token/README.md
+++ b/registry/coder/modules/vault-token/README.md
@@ -20,11 +20,12 @@ variable "vault_token" {
 }
 
 module "vault" {
-  source      = "registry.coder.com/coder/vault-token/coder"
-  version     = "1.1.0"
-  agent_id    = coder_agent.example.id
-  vault_token = var.token # optional
-  vault_addr  = "https://vault.example.com"
+  source          = "registry.coder.com/coder/vault-token/coder"
+  version         = "1.1.0"
+  agent_id        = coder_agent.example.id
+  vault_token     = var.token # optional
+  vault_addr      = "https://vault.example.com"
+  vault_namespace = "prod" # optional, vault enterprise only
 }
 ```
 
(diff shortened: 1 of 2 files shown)


moo-im-a-cow:
the ci check appears to be failing on a different module not modified in this commit, filebrowser,

but i dont see vault-token mentioned in the log so i think the checks stopped at the errors and havent run against the vault-token module yet

DevelopmentCats:
Looks good to me, but can you make sure to update the version in the README to reflect the minor version update to v1.1.1?

registry/coder/modules/vault-token/README.md
```diff
@@ -20,11 +20,12 @@ variable "vault_token" {
...
-  version     = "1.1.0"
-  agent_id    = coder_agent.example.id
-  vault_token = var.token # optional
-  vault_addr  = "https://vault.example.com"
+  source          = "registry.coder.com/coder/vault-token/coder"
+  version         = "1.1.1"
+  agent_id        = coder_agent.example.id
+  vault_token     = var.token # optional
+  vault_addr      = "https://vault.example.com"
+  vault_namespace = "prod" # optional, vault enterprise only
```
DevelopmentCats, line 28: Since this is a minor version update you would actually want to bring this to 1.2.0 since you are adding an input. @moo-im-a-cow

PR #110, opened 2025-05-19
fix: add build step for deployment registry

Parkreiner:
## Changes made
- Updated the `deploy-registry` script to include steps for pushing to the production registry


bcpeinhardt (approved):
😎

PR #111, opened 2025-05-19
chore: update README tags for clarity and consistency

matifali:
- reduced excessive use of `helper`

### DIFF
--- a/registry/nataindata/modules/apache-airflow/README.md
+++ b/registry/nataindata/modules/apache-airflow/README.md
@@ -3,9 +3,8 @@ display_name: airflow
 description: A module that adds Apache Airflow in your Coder template
 icon: ../../../../.icons/airflow.svg
 maintainer_github: nataindata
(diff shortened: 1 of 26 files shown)


registry/nataindata/modules/apache-airflow/README.md
```diff
@@ -3,9 +3,8 @@ display_name: airflow
 description: A module that adds Apache Airflow in your Coder template
 icon: ../../../../.icons/airflow.svg
 maintainer_github: nataindata
-partner_github: coder
```
matifali, line 6: This no longer works.

DevelopmentCats:
@matifali Do we need to update versions on these modules as well?

matifali:
> @matifali Do we need to update versions on these modules as well?

No. We can update frontend without pushing new tags.

DevelopmentCats (approved):

PR #123, opened 2025-05-29
feat: add group attributes to all modules

aslilac:
Goes along with https://github.com/coder/coder/issues/8237

Most people probably gets apps from modules, and so to group them we'll need new versions of aaaaaall of these modules.

Also some were missing `order`, which is related and intertwined pretty closely in the implementation of `group`, so add it where necessary.

### DIFF
--- a/examples/modules/README.md
+++ b/examples/modules/README.md
@@ -15,7 +15,7 @@ tags: [helper]
 module "MODULE_NAME" {
   count   = data.coder_workspace.me.start_count
   source  = "registry.coder.com/NAMESPACE/MODULE_NAME/coder"
-  version = "1.0.2"
+  version = "1.0.0"
 }
 ```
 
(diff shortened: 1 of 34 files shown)


matifali (changes requested):
Also need to bump the minimum required provider version to >= 2.5.3.

The latest related [commit](https://github.com/coder/terraform-provider-coder/commit/cfa101df4635e405e66094fa7779f9a89d92f400) to be tagged.


Can you also bump minor version of each of these modules in README.md?

bcpeinhardt (changes requested):
I dig it 😎 Do you mind going ahead and updating the versions appropriately in the module readmes?

bcpeinhardt:
@aslilac looks like Aider and KasmVNC still need version updates in the readmes.

matifali (changes requested):
Bump minor version please. We should have a script to do that. 
`update_version.sh major|minor|patch`

matifali (approved):
Thank you.

bcpeinhardt (approved):

PR #124, opened 2025-05-30
fix(claude-code): create folder if it does not exist

hugodutka:


kylecarbs (approved):

matifali:
We needed to update the patch version here.

PR #135, opened 2025-06-03
feat: add claude_code_web in claude-code

BrunoQuaresma:

PR #136, opened 2025-06-05
Query agentapi for claude-code status

code-asher:
Depends on https://github.com/coder/registry/pull/135 and https://github.com/coder/coder/pull/18262

~As discussed, this adds a bash script that gets the status from agentapi then relays that to coderd. I wonder if we should make it part of `coder exp` though?  Handling JSON in Bash is pretty janky.  Or we could do it in JS since we already have Node available.~

Refactored into `coder exp`.

~In any case, it looks like `/status` just reports `stable` or `running`. Do we need to modify this endpoint or is there something else I should be using? Initially I thought the endpoint we need already existed but I may have misunderstood.~

We decided to just update the status and preserve the last message if any for now.

### DIFF
--- a/registry/coder/modules/claude-code/main.tf
+++ b/registry/coder/modules/claude-code/main.tf
@@ -138,7 +138,7 @@ resource "coder_script" "claude_code" {
 
     if [ "${var.experiment_report_tasks}" = "true" ]; then
       echo "Configuring Claude Code to report tasks via Coder MCP..."
-      coder exp mcp configure claude-code ${var.folder}
+      coder exp mcp configure claude-code ${var.folder} --ai-agentapi-url http://localhost:3284
     fi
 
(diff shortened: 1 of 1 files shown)


code-asher:
Ahahaha I was wondering what was going on lol

hugodutka:
> Do we need to modify this endpoint or is there something else I should be using

In my mind you'd poll agentapi for current messages and status and then use an LLM to summarize what it's done since the previous update. You're right - bash is probably not the right tool for this job. What do you think about building this tool into coder? You could spawn it in the workspace with `coder exp task status`. I built a Go SDK that might be useful: https://github.com/coder/agentapi-sdk-go. Coderd also already has access to LLM api keys if the admin supplies them: https://github.com/coder/coder/blob/main/cli/server.go#L2697

PR #137, opened 2025-06-05
feat: add GitHub Actions workflow for automatic version bumping

DevelopmentCats:
# Version Bump Workflow

Automates module version bumping when PR labels are applied.

## Usage

Add one of these labels to a PR:
- `version:patch` - Bug fixes (1.2.3 → 1.2.4)
- `version:minor` - New features (1.2.3 → 1.3.0)  
- `version:major` - Breaking changes (1.2.3 → 2.0.0)

## What it does

1. **Detects modified modules** from PR changes
2. **Gets current version** from latest release tag or README
3. **Calculates new version** based on label
4. **Updates README versions** in module documentation
5. **Commits changes** to the PR
6. **Comments summary** on the PR

## Example

If a PR modifies `registry/coder/git-clone/` and has label `version:patch`:

```
Current: release/coder/git-clone/v1.0.18
New: v1.0.19
Updates: All versions for git-clone module = "1.0.19" in README
```

## Version Detection

- **Tagged modules**: Uses latest `release/namespace/module/vX.Y.Z` tag
- **Untagged modules**: Extracts version from README `version = "X.Y.Z"` 
- **New modules**: Start from v1.0.0

## Notes

- Only updates READMEs that contain version references
- Warns about modules missing proper git tags
- Requires write permissions on repository

### DIFF
--- a/.github/workflows/version-bump.yaml
+++ b/.github/workflows/version-bump.yaml
@@ -1,0 +1,221 @@
+name: Version Bump
+
+on:
+  pull_request:
+    types: [labeled]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  version-bump:
+    if: github.event.label.name == 'version:patch' || github.event.label.name == 'version:minor' || github.event.label.name == 'version:major'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


matifali (changes requested):
Can you extract this to a script in `./scripts/update_veraion.sh` that we can reuse in GitHub workflow?

Also the script should be able to run locally and bump versions for individual or multiple modules.

DevelopmentCats:
Script exists.

Is there a way we can test this workflow in an existing PR before Pushing?

I ran some simulated tests that worked fine, but I haven't run this in an actual GH Workflow so i'm not entirely sure if this works as intended in the context of being run as a workflow.

DevelopmentCats:
Let me know if you are good with merging this now.

matifali (changes requested):
Instead of committing opt for failing CI and making it a required check.

PR #143, opened 2025-06-09
fix: replace broken multi-template PR system with unified template

DevelopmentCats:
Unifies the broken Multi-PR Template I introduced.


DevelopmentCats:
Wanna take a look to see if I missed anything? @matifali

matifali (approved):

PR #144, opened 2025-06-09
feat: add templates and update icon paths

bcpeinhardt:
This PR copies the templates in coder/coder/examples/templates over to the registry, so that template contribution can be done through the registry.
For now, the starter templates in the coder/coder binary and the templates available in coder/registry will simply be different constructs, until we find a solution we like around a single source of truth for templates that doesn't raise hairy semver concerns for coder/coder: https://codercom.slack.com/archives/C05T7165ET1/p1749493368773469


registry/coder/templates/aws-devcontainer/architecture.svg
bcpeinhardt, line 1: Under Devcontainer, see "filesytem" is now "filesystem"

.icons/do.png
matifali, line 1: Maybe we replace all icons with SVGs, where possible?

.icons/1f4e6.png
matifali, line 1: Renaming this to something like `box.png` would help avoid any future confusion.

bcpeinhardt: Can do, this was the name it had in coder/coder so I just stuck with it.

.icons/do.png
matifali: Yeah, we generally opt for SVG icons where available for all new icons on `coder/coder` too.

matifali: We should find the original SVGs instead of converting.

PR #172, opened 2025-07-02
chore(modules/claude-code): update  readme

hugodutka:
The README points to an old version of claude code that doesn't support the `--continue` flag, making it incompatible with the latest version of the module. It's easy to copy and paste the example and run into an error. This PR updates the example version and adds troubleshooting notes.


matifali:
@bcpeinhardt do we still support updating the README without pushing a new version tag?

matifali (approved):
LGTM.

bcpeinhardt:
> @bcpeinhardt do we still support updating the README without pushing a new version tag?

No we build the front end content entirely from the latest release now, that was the trade off we made to avoid the front end content being more up to date than the back end.

PR #177, opened 2025-07-03
feat: add the agentapi module

hugodutka:
The agentapi module is a building block for modules that need to run an agentapi server. It's not meant for end users of Coder.

The agentapi-specific logic is mostly extracted from [the claude-code module](https://github.com/coder/registry/tree/c0f2d945c5808c1962b15b5c2e2a5269ea0f4339/registry/coder/modules/claude-code). You can see this module in action in [the goose 2.0 PR](https://github.com/coder/registry/pull/178).

### DIFF
--- a/registry/coder/modules/agentapi/main.tf
+++ b/registry/coder/modules/agentapi/main.tf
@@ -1,0 +1,213 @@
+terraform {
+  required_version = ">= 1.0"
+
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = ">= 2.7"
+    }
+  }
+}
+
+variable "agent_id" {
+  type        = string
+  description = "The ID of a Coder agent."
+}
+
+data "coder_workspace" "me" {}
+
+data "coder_workspace_owner" "me" {}
+
+variable "web_app_order" {
+  type        = number
+  description = "The order determines the position of app in the UI presentation. The lowest order is shown first and apps with equal order are sorted by name (ascending order)."
+  default     = null
+}
+
+variable "web_app_group" {
+  type        = string
+  description = "The name of a group that this app belongs to."
+  default     = null
+}
+
+variable "web_app_icon" {
+  type        = string
+  description = "The icon to use for the app."
+}
+
+variable "web_app_display_name" {
+  type        = string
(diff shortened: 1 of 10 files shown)


registry/coder/modules/agentapi/main.tf
```diff
@@ -0,0 +1,277 @@
...
+variable "install_agentapi" {
+  type        = bool
+  description = "Whether to install AgentAPI."
+  default     = true
+}
+
+variable "agentapi_version" {
+  type        = string
+  description = "The version of AgentAPI to install."
+  default     = "v0.2.3"
```
matifali, line 120: Is there a way to make this `latest` and fetch the latest release version?
We have a similar logic in `vault-github` module.

https://github.com/coder/registry/blob/c0f2d945c5808c1962b15b5c2e2a5269ea0f4339/registry/coder/modules/vault-github/main.tf#L35-L43

registry/coder/modules/agentapi/main.tf
```diff
@@ -0,0 +1,277 @@
...
+      "$module_path/post_install.sh" 2>&1 | tee "$module_path/post_install.log"
+    fi
+
+    export LANG=en_US.UTF-8
+    export LC_ALL=en_US.UTF-8
+
+    cd "${local.workdir}"
+    nohup "$module_path/scripts/agentapi-start.sh" true "${var.agentapi_port}" &> "$module_path/agentapi-start.log" &
+    "$module_path/scripts/agentapi-wait-for-start.sh" "${var.agentapi_port}"
+    EOT
```
matifali, line 233: I suggest moving this to a `script.sh` file and using `templatefile()` function. It will make maintenance easy.

matifali (commented):
Left a few suggestions. BUt LGTM

registry/coder/modules/agentapi/main.tf
hugodutka: Yes - I'd like to avoid that. I don't want `latest` to be the default value for the reasons I stated above.

matifali (approved):
Approving as its an internal module. Can test when being used in other modules.

PR #191, opened 2025-07-07
fix: update namespace for coder-labs template

Parkreiner:
## Description

This PR updates the `maintainer_github` field for the new Coder Labs template to use the value of `coder-labs`. This shouldn't be necessary (`maintainer_github` should be deprecated), but something is off with our build step.

## Type of Change

- [ ] New module
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [x] Other

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

---

## Related Issues

None

### DIFF
--- a/registry/coder-labs/templates/tasks-docker/README.md
+++ b/registry/coder-labs/templates/tasks-docker/README.md
@@ -2,7 +2,7 @@
 display_name: Tasks on Docker
 description: Run Coder Tasks on Docker with an example application
 icon: ../../../../.icons/tasks.svg
-maintainer_github: coder
+maintainer_github: coder-labs
 verified: false
 tags: [docker, container, ai, tasks]
 ---


registry/coder-labs/templates/tasks-docker/README.md
```diff
@@ -2,7 +2,7 @@
 display_name: Tasks on Docker
 description: Run Coder Tasks on Docker with an example application
 icon: ../../../../.icons/tasks.svg
-maintainer_github: coder
+maintainer_github: coder-labs
```
matifali, line 5: Should we instead remove this line? we do not have a `coder-labs` github org

Both `coder` and `coder-labs` have the same GitHub account, `coder`.

bpmct: I just made https://github.com/coder-contrib more official.

I am confused why this is being used as the namespace and not the folder name under registry/

matifali: That's a good workaround for now. I am not sure though if it work. We have yet to test. workflow where `namepsace` and the `maintainer_github` are different.

Parkreiner: I legitimately think that keeping the `maintainer_github` field, and just swapping the value to `coder-labs` is the least destructive way we can update things right now

Parkreiner: Then I can spend today actually getting something in place to make the fields unnecessary (even though it should've been in place back in April/May)

matifali: Ok. Lets do this then to get the template out. And we can fix the build process meanwhile.

matifali (approved):
A workaround to put the template in the correct namespace. We need to fix the build process.

PR #264, opened 2025-07-30
chore: deploy registry changes nightly

### LINKED ISSUE #133: Nightly Rebuild
We want to implement a nightly rebuild for the registry, which will keep things like download counts and the last test run status relatively up to date. This can be done simply by adding a cron specification to https://github.com/coder/registry/blob/main/.github/workflows/deploy-registry.yaml.

matifali:
Closes #133

### DIFF
--- a/.github/workflows/deploy-registry.yaml
+++ b/.github/workflows/deploy-registry.yaml
@@ -1,6 +1,9 @@
 name: deploy-registry
 
 on:
+  schedule:
+    # Runs at 02:30 UTC Monday through Friday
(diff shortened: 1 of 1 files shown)


bcpeinhardt:
I suppose this makes sense for the templates and contributor pages that we may want released without a release tag 😎

bcpeinhardt (approved):

matifali:
@bcpeinhardt Yes. We are already doing a release on each merge to `main` now. This will mainly benefit download count updates, even if there are no new contributions to the registry.

PR #272, opened 2025-07-31
chore: polish some modules readme

phorcys420:


DevelopmentCats (changes requested):
The only thing that I notice is that we should update the versions in the README since this will require a patch release for each module.

You will also want to run `bun fmt` on this so the CI will pass

matifali (approved):

PR #278, opened 2025-07-31
feat: add `vscode-desktop-core` module

phorcys420:


registry/coder/modules/vscode-desktop-core/README.md
```diff
@@ -0,0 +1,33 @@
...
+  source  = "registry.coder.com/coder/vscode-desktop-core/coder"
+  version = "1.0.1"
+
+  agent_id = var.agent_id
+
+  web_app_icon         = "/icon/code.svg"
+  web_app_slug         = "vscode"
+  web_app_display_name = "VS Code Desktop"
+  web_app_order        = var.order
+  web_app_group        = var.group
```
matifali, line 27:
```suggestion
  coder_app_icon         = "/icon/code.svg"
  coder_app_slug         = "vscode"
  coder_app_display_name = "VS Code Desktop"
  coder_app_order        = var.order
  coder_app_group        = var.group
```

I would prefer calling these `coder_app_*` as these are for actually an external coder_app that's doesn't open in your browser.

registry/coder/modules/vscode-desktop-core/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  default     = ""
+}
+
+variable "open_recent" {
+  type        = bool
+  description = "Open the most recent workspace or folder. Falls back to the folder if there is no recent workspace or folder to open."
+  default     = false
+}
+
+variable "protocol" {
```
matifali, line 29: we have a white list of protocols in `coder/coder` to inject the secret `SESSION_TOKEN`. So, maybe we add a validation here to control what we support.

registry/coder/modules/vscode-desktop-core/README.md
phorcys420: @matifali good catch, I only copied what AgentAPI does, but it is indeed not web.
quick thought though, should we also change AgentAPI?

registry/coder/modules/vscode-desktop-core/main.tf
phorcys420: hmm i'll find that list

registry/coder/modules/vscode-desktop-core/README.md
matifali: Agent API does add a webapp and is only browser only.

phorcys420: yeah that is right, but it's still a coder_app though, just said that so we could maybe make it more coherent between modules

registry/coder/modules/vscode-desktop-core/main.tf
matifali: It wasn't asking when I was doing the kiro module.

matifali: But yeah we can skip validation

phorcys420: I will just add a note then

registry/coder/modules/vscode-desktop-core/README.md
```diff
@@ -0,0 +1,33 @@
...
+verified: true
+tags: [internal, library]
+---
+
+# VS Code Desktop Core
+
+> [!CAUTION]
+> We do not recommend using this module directly. Instead, please consider using one of our [Desktop IDE modules](https://registry.coder.com/modules?search=tag%3Aide).
+
+The VSCode Desktop Core module is a building block for modules that need to expose access to VSCode-based IDEs. It is intended primarily for internal use by Coder to create modules for VSCode-based IDEs.
```
matifali, line 14:
```suggestion
The VSCode Desktop Core module is a building block for modules that need to expose access to VSCode-based IDEs. It is intended primarily to be used as a library to create modules for VSCode-based IDEs.
```

Let's make clear that anyone can use it to support any future VS Code-based IDE modules.

matifali (approved):
Thanks a lot @phorcys420. This will significantly enhance the DevEX for module development and maintenance.

matifali:
Thoughts on adding MCP server support?
We can add a generic input named `mcp_config` and `mcp_path`, which we can use from browser modules. 

Here are a few examples
1. [Windsurf](https://docs.windsurf.com/windsurf/cascade/mcp): `/Users/matifali/.codeium/windsurf/mcp_config.json`
2. [Cursor](https://docs.cursor.com/en/context/mcp#configuration-locations): `~/.cursor/mcp.json`
3. [VS Code](https://code.visualstudio.com/docs/copilot/chat/mcp-servers#_add-an-mcp-server): Didn't find a user specfic setting path

matifali:
Nice. But I think exposing MCPs separately would make more sense given how popular they are.

PR #279, opened 2025-07-31
feat: update vscode-based desktop IDE modules to use `vscode-desktop-core`

phorcys420:
Depends on #278

Part of WL-27


matifali:
Does it cover all of them? We recently added positorn and antigravity modules.

PR #287, opened 2025-08-04
fix(devcontainers-cli): allow yarn to install when `packageManager` not `yarn`

DanielleMaywood:
## Description

On our dogfood workspaces, we fail to install `@devcontainers/cli` with `yarn` because our agent directory `/home/coder/coder` contains a `package.json` with `packageManager` being set to `pnpm`. This change instead ensures to run `yarn global add` inside the `$CODER_SCRIPT_DATA_DIR` so that we don't read a `package.json` and cause things to break.

## Type of Change

- [ ] New module
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/devcontainers-cli`  
**New version:** `v1.0.32`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

None

### DIFF
--- a/registry/coder/modules/devcontainers-cli/run.sh
+++ b/registry/coder/modules/devcontainers-cli/run.sh
@@ -38,7 +38,10 @@ install() {
         fi
         pnpm add -g @devcontainers/cli
     elif [ "$PACKAGE_MANAGER" = "yarn" ]; then
-        yarn global add @devcontainers/cli --prefix "$(dirname "$CODER_SCRIPT_BIN_DIR")"
+        # We want to cd into `$CODER_SCRIPT_DATA_DIR` as the current directory
+        # might contain a `package.json` with `packageManager` set to something
+        # other than `yarn`. When this happens, `yarn global add` will fail to install.
+        cd "$CODER_SCRIPT_DATA_DIR" && yarn global add @devcontainers/cli --prefix "$(dirname "$CODER_SCRIPT_BIN_DIR")"
     fi
 }
 
--- a/registry/coder/modules/devcontainers-cli/README.md
+++ b/registry/coder/modules/devcontainers-cli/README.md
@@ -15,7 +15,7 @@ The devcontainers-cli module provides an easy way to install [`@devcontainers/cl
 ```tf
 module "devcontainers-cli" {
   source   = "registry.coder.com/coder/devcontainers-cli/coder"
-  version  = "1.0.31"
+  version  = "1.0.32"
   agent_id = coder_agent.example.id
 }
 ```
--- a/registry/coder/modules/devcontainers-cli/main.test.ts
+++ b/registry/coder/modules/devcontainers-cli/main.test.ts
@@ -45,12 +45,16 @@ const executeScriptInContainerWithPackageManager = async (
 
   console.log(path);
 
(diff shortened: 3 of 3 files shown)


registry/coder/modules/devcontainers-cli/run.sh
```diff
@@ -38,7 +38,10 @@ install() {
         fi
```
mafredri, line 38: Drive by: noticed we export `M_HOME` here lol.

registry/coder/modules/devcontainers-cli/run.sh
```diff
@@ -38,7 +38,10 @@ install() {
...
         pnpm add -g @devcontainers/cli
     elif [ "$PACKAGE_MANAGER" = "yarn" ]; then
-        yarn global add @devcontainers/cli --prefix "$(dirname "$CODER_SCRIPT_BIN_DIR")"
+        # We want to cd into `$CODER_SCRIPT_DATA_DIR` as the current directory
+        # might contain a `package.json` with `packageManager` set to something
+        # other than `yarn`. When this happens, `yarn global add` will fail to install.
+        cd "$CODER_SCRIPT_DATA_DIR" && yarn global add @devcontainers/cli --prefix "$(dirname "$CODER_SCRIPT_BIN_DIR")"
```
mafredri, line 44: Consider using `pushd "..." && yarn ... && popd` to ensure we don't change the state of the script after install (as-is: pre-install cwd = something, after install cwd = data dir).

Alternatively, we could move the `cd` to the top of the script so that the assumption is we remain there for the rest of the script.

mafredri (approved):
Thanks for fixing! Approved, but I'd suggest changing the `cd` slightly to avoid side-effects.

registry/coder/modules/devcontainers-cli/run.sh
DanielleMaywood: Huh, lol, wonder if that was meant to be `PNPM_HOME`

registry/coder/modules/devcontainers-cli/run.sh
DanielleMaywood: I've made a change to use `pushd` and `popd` instead, but this has required replacing our usage of posix `sh` with `bash`. I think replacing the shell used _may_ require us to do a major version bump as it is a breaking change for users that do not have `bash` installed.

mafredri: True, forgot about them being internal functions. Then perhaps we should retain `sh` compat? I know most scripts have abandoned it in favor of Bashisms but I have a soft-spot for it. My two alternative options are:

1. `cd` early in the script
2. wrap in parenthesis (i.e. `(cd ... && install)`)

Option 2 works because it creates a sub shell and the cwd change only applies to it.

DanielleMaywood: I agree with retaining `sh` compatibility (I also have a soft-spot for it).

I'm fine with either approach you've suggested. I'm happy to go for the `cd` early as it might save us from a similar issue happening with the other installation approaches.

DanielleMaywood: I've gone with the early `cd` approach, and it appears to work with local testing 👍

Parkreiner (approved):

PR #320, opened 2025-08-11
feat(agentapi): use wildcard alllowed hosts

hugodutka:
Since https://github.com/coder/agentapi/pull/49 was merged, agentapi by default only accepts requests with the `Host` header set to localhost, 127.0.0.1, or [::1]. In Coder, agentapi is served behind a reverse proxy as a workspace app, so we need to use a wildcard `AGENTAPI_ALLOWED_HOSTS` for agentapi-based modules to continue working.

This PR updates the claude code and agentapi modules, and a subsequent PR will update modules that are based on the agentapi module.


matifali (approved):
LGTM.

johnstcn (approved):

Issue #194, opened 2025-07-07
Add validation for Template and Module README files


Parkreiner:
This is a follow up to these issues/PRs: 
- https://github.com/coder/registry/pull/144
- https://github.com/coder/registry/pull/182
- https://github.com/coder/registry/pull/188

When we brought in all the templates that were previously in `coder/coder`, we never updated any other files. Notably, we never added validation logic to ensure that all the README files were formatted correctly, or had any required content

That already caused a problem when we tried to publish a template under the new Coder Labs namespace.

We can definitely reuse a good amount of the validation logic we have right now. The main thing we need to figure out is, beyond basic README structure validation, is there any specific content that template READMEs should have? For example, module READMEs require an example Terraform snippet so that users can easily bring them in. Should templates have anything similar?

## Action steps
1. Identify what template-specific content (if any) we want to require from template README files
2. Update validation logic to ensure that templates follow base guidelines for README formatting and have all required content
3. Update any template files as necessary to match new validation requirements

PR #325, opened 2025-08-13
fix: improve logic for existing README validation

buenos-nachos:
Addresses part of https://github.com/coder/registry/issues/194

## Description

This PR beefs up the validation for the validation logic that we already had in place. This PR does not include adding validation for templates (which will be addressed in a second PR).

### Changes made
- Added logic to reject unknown frontmatter fields for modules and contributor profile README files
- Added logic to handle frontmatter fields that were previously missed in validation steps (GitHub username for contributors and Operating Systems for modules)
- Updated a few comments (added some new comments, formatted existing comments to meet 100-column width)

### Type of Change

- [ ] New module
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [x] Other

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

### DIFF
--- a/cmd/readmevalidation/coderresources.go
+++ b/cmd/readmevalidation/coderresources.go
@@ -38,7 +38,7 @@ type coderResourceFrontmatter struct {
 // list at runtime in the future, but this should be okay for now
 var supportedCoderResourceStructKeys = []string{
 	"description", "icon", "display_name", "verified", "tags", "supported_os",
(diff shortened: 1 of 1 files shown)


DevelopmentCats (approved):
Looks Great! Ill wait for Atif to look at this before we merge it just in case he has some thoughts on other things we could validate here outside of template stuff like you mentioned.

PR #388, opened 2025-08-26
fix(ci): add fallback for GitHub API failures in release workflow

DevelopmentCats:
## Description

CI was failing on new module releases because there was no fallback to gh api failures when there was no previous tag for the module was found.

https://github.com/coder/registry/actions/runs/17225186737/job/48868318539

## Type of Change

- [ ] New module
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [X] Other

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,13 +73,15 @@ jobs:
             exit 0
           fi
           
-          FULL_CHANGELOG=$(gh api repos/:owner/:repo/releases/generate-notes \
-            --field tag_name="$CURRENT_TAG" \
-            --field previous_tag_name="$PREV_TAG" \
-            --jq '.body' 2>/dev/null) || {
-            echo "GitHub API failed (likely new module), proceeding with git-based changelog"
+          if [[ "$PREV_TAG" == release/* ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -76,7 +76,10 @@ jobs:
           FULL_CHANGELOG=$(gh api repos/:owner/:repo/releases/generate-notes \
             --field tag_name="$CURRENT_TAG" \
             --field previous_tag_name="$PREV_TAG" \
-            --jq '.body')
+            --jq '.body' 2>/dev/null) || {
+            echo "GitHub API failed (likely new module), proceeding with git-based changelog"
+            FULL_CHANGELOG=$(git log --pretty=format:"* %s by %an" --no-merges "$PREV_TAG..$CURRENT_TAG" -- "$MODULE_PATH")
```
matifali, line 81: Look alike this will always overwrite what the previous step does.

matifali (commented):
As a test, we can try creating release for all existing modules manually.

.github/workflows/release.yml
DevelopmentCats: oh well that is dumb, we definitely don't want that. I figured when it 404'd it would fallback rather than falling back always.

matifali (approved):

PR #439, opened 2025-09-29
Fix claude-code module not passing workdir to agentapi

### LINKED ISSUE #436: claude-code 3.0.0 doesn't work with a custom workdir
The  claude-code 3.0.0 module uses the agentapi module.  Setting claude-code variable `workdir` doesn't work because `workdir` is not passed to agentapi`folder` variable, which will [default](https://github.com/coder/registry/blob/e516446d03fd39dc96b50b1da0410745a38c92fa/registry/coder/modules/agentapi/main.tf#L51.
) to `/home/coder` 

Add this line to https://github.com/coder/registry/blob/e516446d03fd39dc96b50b1da0410745a38c92fa/registry/coder/modules/claude-code/main.tf#L241

```
folder = local.workdir
```

matifali:
## Summary

Fixes #436 - The claude-code 3.0.0 module was not passing the custom `workdir` variable to the agentapi module, causing it to default to `/home/coder` instead of using the specified working directory.

## Changes

- Added missing `folder = local.workdir` parameter to the agentapi module call in `main.tf:247`
- This ensures that custom working directories are properly propagated to the agentapi module

## Test Plan

- [x] Terraform validation passes
- [x] Code formatting applied with `bun run fmt`
- [x] Basic terraform test passes (one pre-existing test failure unrelated to this change)

## Verification

The fix adds the missing parameter that was identified in the issue:
```terraform
module "agentapi" {
  # ... other parameters
  folder = local.workdir  # <- Added this line
  # ... rest of configuration
}
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/registry/coder/modules/claude-code/README.md
+++ b/registry/coder/modules/claude-code/README.md
@@ -13,7 +13,7 @@ Run the [Claude Code](https://docs.anthropic.com/en/docs/agents-and-tools/claude
 ```tf
 module "claude-code" {
   source         = "registry.coder.com/coder/claude-code/coder"
-  version        = "3.0.0"
+  version        = "3.0.1"
   agent_id       = coder_agent.example.id
   workdir        = "/home/coder/project"
   claude_api_key = "xxxx-xxxxx-xxxx"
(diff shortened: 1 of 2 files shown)


DevelopmentCats:
Im not too sure how this wasn't passed especially since we have the local defined for workdir.

Thanks Atif! I Noticed you requested review, but this is still in draft.

PR #443, opened 2025-10-01
feat(claude-code): add coder-specific prompt to system_prompt

ssncferreira:
## Description

This PR updates the `claude-code` module to automatically include the Coder task-reporting system prompt whenever `report_tasks = true`, and to wrap the final system prompt in `<system>…</system>` when non-empty. 

Previously, users needed to manually include this content in their system prompts to enable proper task reporting. When `report_tasks = true`, the system prompt is prepended with the Coder task-reporting, and any user `system_prompt` (if provided) is appended after it, ensuring consistent integration without manual copy/paste.

When `report_tasks = false`, the module includes only the user `system_prompt` (if any). If both `report_tasks = false` and `system_prompt` is empty, the system prompt sent to Claude is empty. 

## Type of Change

- [ ] New module
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v3.0.2` 
**Breaking change:** [] Yes [x]  No

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

Related to internal slack thread: https://codercom.slack.com/archives/C0992H8HGCS/p1759317555713269

### DIFF
--- a/registry/coder/modules/claude-code/scripts/install.sh
+++ b/registry/coder/modules/claude-code/scripts/install.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -euo pipefail
 
-source "$HOME"/.bashrc
+if [ -f "$HOME/.bashrc" ]; then
+  source "$HOME"/.bashrc
+fi
 
 BOLD='\033[0;1m'
 
--- a/registry/coder/modules/claude-code/scripts/start.sh
+++ b/registry/coder/modules/claude-code/scripts/start.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -euo pipefail
 
-source "$HOME"/.bashrc
+if [ -f "$HOME/.bashrc" ]; then
+  source "$HOME"/.bashrc
+fi
 export PATH="$HOME/.local/bin:$PATH"
 
 command_exists() {


registry/coder/modules/claude-code/main.tf
```diff
@@ -182,8 +182,8 @@ variable "claude_code_oauth_token" {
 
 variable "system_prompt" {
   type        = string
-  description = "The system prompt to use for the Claude Code server."
-  default     = "Send a task status update to notify the user that you are ready for input, and then wait for user input."
+  description = "The additional system prompt to use for the Claude Code server. The module includes mandatory sections inside a <system> block. Leave empty to include only the mandatory sections."
+  default     = ""
```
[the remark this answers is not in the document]
mafredri: Do we foresee applying user prompt magic as well? If so, a single option like `use_coder_prompt_magic` (or `disable_*`) might be a good fit. A bit worried about having too many parameters.

ssncferreira:
@johnstcn @mafredri @DevelopmentCats thank you for the review.
What if we keep `system_prompt` as-is and introduce a new boolean variable `include_coder_system_prompt` that defaults to `true`? Meaning that:
* if `include_coder_system_prompt = true`, we append whatever the user specifies in `system_prompt`, if any.
* if `include_coder_system_prompt = false`, `system_prompt` fully replaces the prompt.

This would still be a breaking change, and we would need to update the documentation to recommend `include_coder_system_prompt = true` for proper integration with Coder. Wdyt?

ssncferreira:
> My understanding is that the "breaking" part of this change would be defaulting `include_coder_system_prompt` to `true`. We could have a cutover period where we default this to `false` initially (current behaviour) and recommend that this is set to `true` in the documentation. We may decide to default this to `true` later. WDYT?

That’s a good approach. We can release a minor version now with `include_coder_prompt_sections` added (default to `false`) so users can opt in, and then change the default to `true` in the next major release 👍

DevelopmentCats:
> > My understanding is that the "breaking" part of this change would be defaulting `include_coder_system_prompt` to `true`. We could have a cutover period where we default this to `false` initially (current behaviour) and recommend that this is set to `true` in the documentation. We may decide to default this to `true` later. WDYT?
> 
> That’s a good approach. We can release a minor version now with `include_coder_system_prompt` added (default to `false`) so users can opt in, and then change the default to `true` in the next major release 👍

Sounds good to me! Once this is done I'll this to the list of things to expand to the other AI modules as well.

registry/coder/modules/claude-code/main.tf
```diff
@@ -231,6 +236,42 @@ locals {
   start_script                      = file("${path.module}/scripts/start.sh")
   module_dir_name                   = ".claude-module"
   remove_last_session_id_script_b64 = base64encode(file("${path.module}/scripts/remove-last-session-id.sh"))
+
+  # Required prompts for the module to properly integrate with Coder
+  inner_system_prompt = <<-EOT
+      -- Tool Selection --
```
ssncferreira, line 242: I included only the `-- Tool Selection --` and `-- Task Reporting --` sections since they seem to be the only required for reliable Coder integration. In our dogfood we also have a `-- Framing --` and `-- Context --`, but those feel more org-specific, so I left them out. Let me know if you think it also makes sense to add those here as well

registry/coder/modules/claude-code/main.tf
```diff
@@ -231,6 +236,42 @@ locals {
...
+      - coder_report_task: providing status updates or requesting user input.
+      - playwright: previewing your changes after you made them
+        to confirm it worked as expected
+      -	desktop-commander - use only for commands that keep running
+        (servers, dev watchers, GUI apps).
```
johnstcn, line 247: These MCP servers aren't guaranteed to be present.

registry/coder/modules/claude-code/main.tf
```diff
@@ -231,6 +236,42 @@ locals {
   start_script                      = file("${path.module}/scripts/start.sh")
   module_dir_name                   = ".claude-module"
   remove_last_session_id_script_b64 = base64encode(file("${path.module}/scripts/remove-last-session-id.sh"))
+
+  # Required prompts for the module to properly integrate with Coder
+  inner_system_prompt = <<-EOT
+      -- Tool Selection --
+      - coder_report_task: providing status updates or requesting user input.
```
johnstcn, line 243: This will only be present if task reporting is enabled

ssncferreira: Does that mean that this should only be present is variable `report_tasks` is true?

johnstcn: That's my understanding of the module, yes.

ssncferreira: Addressed as part of: https://github.com/coder/registry/pull/443/commits/592e40c227ef23eb2f49e91c785e67df2d02284d

Originally, `include_coder_system_prompt` was intended to cover all system prompts, but because the injected prompt is now specific to task reporting, the variable has been renamed to `report_tasks_system_prompt`. Because this behavior is tied to `report_tasks`, and in order to avoid having a dedicated system prompt toggle for every feature, it may make sense to consolidate into a single structured setting in a future major release. For the next major, this could look like:
```
variable "coder_task_reporting" {
  description = "Task reporting configuration. Enables AgentAPI reporting and injects the Coder task-reporting prompt."
  type = object({
    enabled        = optional(bool, true)
    include_prompt = optional(bool, true) 
  })
  default = {}  # both default to true
}
```

johnstcn (approved):

registry/coder/modules/claude-code/README.md
```diff
@@ -26,6 +26,14 @@ module "claude-code" {
 > [!NOTE]
 > By default, this module is configured to run the embedded chat interface as a path-based application. In production, we recommend that you configure a [wildcard access URL](https://coder.com/docs/admin/setup#wildcard-access-url) and set `subdomain = true`. See [here](https://coder.com/docs/tutorials/best-practices/security-best-practices#disable-path-based-apps) for more details.
 
+<!-- TODO(major): remove this note and update the prompt configuration example in the next major release -->
```
mafredri, line 29: Do we have an issue to track these TODOs? If not, they are easily overlooked.

mafredri (approved):
LGTM, thanks for implementing this change @ssncferreira!

registry/coder/modules/claude-code/README.md
ssncferreira: I don't think so, I tried adding as much information as possible in the TODOs to avoid having this overlooked...but not sure how to enforce it 😕

registry/coder/modules/claude-code/main.tf
```diff
@@ -186,6 +186,15 @@ variable "system_prompt" {
...
 }
 
+variable "report_tasks_system_prompt" {
+  type        = bool
+  description = "Include the Coder system prompt for task reporting when report_tasks is true. Defaults to false for backward compatibility, will default to true in the next major release."
+  # TODO(major): default to true in the next major release
+  #   Context PR: https://github.com/coder/registry/pull/443
+  #   See discussion for the `report_tasks` variable definition: https://github.com/coder/registry/pull/443#discussion_r2405788746
+  default = false
+}
```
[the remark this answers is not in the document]
ssncferreira: That was the initial suggestion, but it was decided that we don't want to enforce this system prompt to users. If they want to set their own system prompt regarding task reporting, they can. See previous PR comments for context, mainly https://github.com/coder/registry/pull/443#discussion_r2395212674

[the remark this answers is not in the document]
matifali: can we do something like

```tf
locals {
	task_reporting_prompt = "TASK REPORTING PROMPT"
	final_system_prompt = join(", [var.system_prompt, var.report_tasks ? " ${var.task_reporting_prompt}" : ""]  
}

```

DevelopmentCats (approved):

registry/coder/modules/claude-code/main.tf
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/registry/pull/443/commits/e0502017ae086e454ff521f8899bf991a6d236f0

registry/coder/modules/claude-code/main.tf
```diff
@@ -231,6 +229,37 @@ locals {
...
+  # Only include coder system prompts if report_tasks is enabled
+  inner_system_prompt = format(
+    "%s%s",
+    var.report_tasks ? format("\n%s\n", local.report_tasks_system_prompt) : "",
+    trimspace(local.custom_system_prompt) != "" ? format("\n%s\n", local.custom_system_prompt) : "",
+  )
+  final_system_prompt = trimspace(local.inner_system_prompt) != "" ? format(
+    "<system>%s</system>",
+    local.inner_system_prompt
+  ) : ""
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/registry/pull/443/commits/9e6dcc8e11e3879b60c0786ef77de95f52ecc431

mafredri (approved):

PR #459, opened 2025-10-06
fix(claude-code): source bashrc file only if it exists

### LINKED ISSUE #458: agentapi server not starting up after upgrade of claude-code module to v3
### Why

Getting following error in coder:

```
Error: agentapi server did not start on port 3284 after 15 seconds.
```

### Possible root cause

- The v3 registry upgrade [refactored the install script](https://github.com/coder/registry/pull/402).
- We don't have a `~/.bashrc` ,so `set -euo pipefail` is causing it to fail

### Workaround

During first 15 seconds:

```
touch ${HOME}/.bashrc
```

schahal:
Closes #458

## Description

It only sources the bashrc file if it exists (as the startup/install scripts will fail if the workspace doesn't have that)

## Type of Change

- [ ] New module
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v3.0.2`   <--- Maintainers, I'll have to lookup how to bump this for this project

## Testing & Validation

- [ ] Tests pass (`bun test`)
- [ ] Code formatted (`bun run fmt`)
- [ ] Changes tested locally

### DIFF
--- a/registry/coder/modules/claude-code/README.md
+++ b/registry/coder/modules/claude-code/README.md
@@ -13,7 +13,7 @@ Run the [Claude Code](https://docs.anthropic.com/en/docs/agents-and-tools/claude
 ```tf
 module "claude-code" {
   source         = "registry.coder.com/coder/claude-code/coder"
-  version        = "3.0.1"
+  version        = "3.0.2"
   agent_id       = coder_agent.example.id
   workdir        = "/home/coder/project"
   claude_api_key = "xxxx-xxxxx-xxxx"
@@ -49,7 +49,7 @@ data "coder_parameter" "ai_prompt" {
 
 module "claude-code" {
   source   = "registry.coder.com/coder/claude-code/coder"
-  version  = "3.0.1"
+  version  = "3.0.2"
   agent_id = coder_agent.example.id
   workdir  = "/home/coder/project"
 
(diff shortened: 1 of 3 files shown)


schahal:
@DevelopmentCats , I'm trying to follow `CONTRIBUTING.md`, but running

```
./.github/scripts/version-bump.sh patch
```

on this branch doesn't return anything:

```
$ ./.github/scripts/version-bump.sh patch
🔍 Detecting modified modules...
$
```

Are there other steps needed, you think?

DevelopmentCats:
> @DevelopmentCats , I'm trying to follow `CONTRIBUTING.md`, but running
> 
> ```
> ./.github/scripts/version-bump.sh patch
> ```
> 
> on this branch doesn't return anything:
> 
> ```
> $ ./.github/scripts/version-bump.sh patch
> 🔍 Detecting modified modules...
> $
> ```
> 
> Are there other steps needed, you think?

You can just manually replace the versions in README if you want. 

This script essentially just updates all the version references in the README to match what they should based on the last tag and requested bump type.

Does the script not exist for you at all in your version?

schahal:
👍 thank you for the quick iterations!

Looks like the PR didn't auto-merge because the branch was out-of-date, so I just updated.

PR #483, opened 2025-10-16
feat: change tf test and validation to use paths-filter

DevelopmentCats:
## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/[namespace]/modules/[module-name]`  
**New version:** `v1.0.0`  
**Breaking change:** [ ] Yes [ ] No

## Template Information

**Path:** `registry/[namespace]/templates/[template-name]`

## Testing & Validation

- [ ] Tests pass (`bun test`)
- [ ] Changes tested locally

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,6 +27,8 @@ jobs:
               - 'tsconfig.json'
               - '.github/workflows/ci.yaml'
               - 'scripts/ts_test_auto.sh'
+              - 'scripts/terraform_test_all.sh'
+              - 'scripts/terraform_validate.sh'
             modules:
               - 'registry/**/modules/**'
             all:


matifali (approved):
LGTM. How much time does this save on CI? Can we do something so that we run all tests if its on `main`?

matifali (approved):

DevelopmentCats:
> LGTM. How much time does this save on CI? Can we do something so that we run all tests if its on `main`?

Realistically it would only save us around 15-60 seconds I think since a bulk of the time is spent on typescript tests when all are run

I've seen a weird variation in ci run time for the validation step in other runs so its hard to estimate

PR #484, opened 2025-10-17
ci: add shellcheck step

DevelopmentCats:


DevelopmentCats:
we should be good here now.

matifali (approved):

PR #496, opened 2025-10-20
fix: resolve issues with claude-code session resumption

DevelopmentCats:
## Description

Fixes session resumption logic by having the continue flag decide whether to continue a workspace based on session history

## Type of Change

- [ ] New module
- [ ] New template
- [X] Bug fix
- [X] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v3.2.2`  
**Breaking change:** [ ] Yes [X] No

## Testing & Validation

- [X] Tests pass (`bun test`)
- [X] Code formatted (`bun fmt`)
- [X] Changes tested locally

### DIFF
--- a/registry/coder-labs/modules/copilot/README.md
+++ b/registry/coder-labs/modules/copilot/README.md
@@ -13,7 +13,7 @@ Run [GitHub Copilot CLI](https://docs.github.com/copilot/concepts/agents/about-c
 ```tf
 module "copilot" {
   source   = "registry.coder.com/coder-labs/copilot/coder"
-  version  = "0.2.1"
+  version  = "0.2.2"
   agent_id = coder_agent.example.id
   workdir  = "/home/coder/projects"
 }
@@ -51,7 +51,7 @@ data "coder_parameter" "ai_prompt" {
 
 module "copilot" {
   source   = "registry.coder.com/coder-labs/copilot/coder"
-  version  = "0.2.1"
+  version  = "0.2.2"
   agent_id = coder_agent.example.id
   workdir  = "/home/coder/projects"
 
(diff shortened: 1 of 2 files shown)


hugodutka (commented):
I see the logic is tightly coupled with the internal file structure of claude. I think it'd be more robust to rely on `claude --session-id <uuid>` and `claude -r <uuid>`.

You could choose a predefined uuid such as `cd32e253-ca16-4fd3-9825-d837e74ae3c2` for the task session. Before starting claude, first run

```sh
find "$HOME/.claude" -type f -name "*cd32e253-ca16-4fd3-9825-d837e74ae3c2*"
```

to see if there are any sessions with this id in the workspace. If there are, use `claude -r`, if there are none, use `--session-id`.

This approach also avoids buggy behavior with user-spawned `claude` sessions in the workspace. `--continue` opens the last-used session, `--session-id` and `-r` would always open the same task session.

DevelopmentCats:
> I see the logic is tightly coupled with the internal file structure of claude. I think it'd be more robust to rely on `claude --session-id <uuid>` and `claude -r <uuid>`.
> 
> You could choose a predefined uuid such as `cd32e253-ca16-4fd3-9825-d837e74ae3c2` for the task session. Before starting claude, first run
> 
> ```shell
> find "$HOME/.claude" -type f -name "*cd32e253-ca16-4fd3-9825-d837e74ae3c2*"
> ```
> 
> to see if there are any sessions with this id in the workspace. If there are, use `claude -r`, if there are none, use `--session-id`.
> 
> This approach also avoids buggy behavior with user-spawned `claude` sessions in the workspace. `--continue` opens the last-used session, `--session-id` and `-r` would always open the same task session.

I ironically just made the same thought and switched over to extracting the latest session with actual user messages in it, but defining the session id is a much better solution. I will make the change now.

PR #501, opened 2025-10-22
chore: Update templates from Always to IfNotPresent for image_pull_policy

rowansmithau:
## Description

Change `image_pull_policy` from `Always` to `IfNotPresent` on Coder owned templates. Given these are a reference point for users and customers and they copy them into their own templates I think it makes sense to encourage the use of caching of images.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Template Information

**Path:** 

https://github.com/coder/registry/tree/main/registry/coder/templates/kubernetes-devcontainer
https://github.com/coder/registry/tree/main/registry/coder/templates/kubernetes-envbox
https://github.com/coder/registry/tree/main/registry/coder/templates/kubernetes

## Testing & Validation

- [ ] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

None

### DIFF
--- a/.github/typos.toml
+++ b/.github/typos.toml
@@ -6,6 +6,7 @@ HashiCorp = "HashiCorp"
 mavrickrishi = "mavrickrishi" # Username
 mavrick = "mavrick" # Username
 inh = "inh" # Option in setpriv command
+exportfs = "exportfs" # nfs related binary
 
 [files]
 extend-exclude = ["registry/coder/templates/aws-devcontainer/architecture.svg"] #False positive
--- a/registry/coder-labs/modules/copilot/README.md
+++ b/registry/coder-labs/modules/copilot/README.md
@@ -13,7 +13,7 @@ Run [GitHub Copilot CLI](https://docs.github.com/copilot/concepts/agents/about-c
 ```tf
 module "copilot" {
   source   = "registry.coder.com/coder-labs/copilot/coder"
-  version  = "0.2.1"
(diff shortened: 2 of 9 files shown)


rowansmithau:
> if someone has build pipeline, where they regularly build an publish new images to a registry, will it not pull the stale image?
> 
> So may be we shoudl add a comment that
> 
> > In CI/CD with frequent image updates, tag images uniquely and keep `IfNotPresent`.
> > If you rely on mutable tags like `latest`, set `image_pull_policy: Always`.

My thought was if someone has a build pipeline they should have a good understanding of their container runtime image pull policy and can set it appropriately for their use case. These are just example getting started templates for which the average Joe will use to begin using Coder, for which I don't feel like it makes sense to have the image pull policy default to `Always`. As it is, `Always` still results in a cached image being reused if present providing the sha/digest has not changed (by default), so this is really just us trying to be providing good guidance to users new to the concept.

rowansmithau:
@matifali it's more around trying to be a good steward of not having excessive / unnecessary API calls be the default behaviour for folks who use our templates / use our templates as a baseline. It would have the added benefit of either matching existing startup times (at worst) or reducing workspace startup time by a few ms/s because it wouldn't be making the call to Docker Hub / ghcr to check the sha (as it does with `Always`).

matifali (approved):

PR #509, opened 2025-10-24
feat: dropping perms before running claude

evgeniy-scherbina:


DevelopmentCats:
@evgeniy-scherbina can you run `./.github/scripts/version-bump.sh patch`

PR #522, opened 2025-10-31
fix: session resumption fix, and bug fixes for arg path logic

35C4n0r:
## Description

Fix issue with commands being injected through prompt.

Bug fix for logic in arg paths.

## Type of Change

- [ ] New module
- [ ] New template
- [X] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v4.0.1`  
**Breaking change:** [ ] Yes [X] No

## Testing & Validation

- [X] Tests pass (`bun test`)
- [X] Code formatted (`bun fmt`)
- [ ] Changes tested locally

### DIFF
--- a/registry/coder-labs/templates/tasks-docker/main.tf
+++ b/registry/coder-labs/templates/tasks-docker/main.tf
@@ -1,7 +1,8 @@
 terraform {
   required_providers {
     coder = {
-      source = "coder/coder"
+      source  = "coder/coder"
+      version = ">= 2.13"
     }
     docker = {
       source = "kreuzwerker/docker"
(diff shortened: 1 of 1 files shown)


DevelopmentCats:
@matifali 

This is good to go

registry/coder/modules/claude-code/scripts/start.sh
```diff
@@ -69,10 +79,15 @@ function validate_claude_installation() {
   fi
 }
 
+# Hardcoded task session ID for Coder task reporting
+# This ensures all task sessions use a consistent, predictable ID
 TASK_SESSION_ID="cd32e253-ca16-4fd3-9825-d837e74ae3c2"
```
35C4n0r: Why are we hardcoding this again, @DevelopmentCats ?

35C4n0r: ref:
```
Chris (Cat)
  [Yesterday at 11:10 PM](https://codercom.slack.com/archives/C09D28669C0/p1762882821216439?thread_ts=1762878980.800449&cid=C09D28669C0)
I think the issue with the continue flag is that the agent could get confused with other sessions that the user has created in the workspace. This is why we decided to go a pre populated session if for tasks so that way the session is always continued properly for tasks.
```

35C4n0r:
LGTM, functionality pov, it doesn't break anymore.

PR #567, opened 2025-11-26
chore: update all module READMEs to use main agent id

phorcys420:
## Summary

This PR standardizes all module documentation examples to use `coder_agent.main` instead of `coder_agent.example`, aligning with the most common template convention.

## Problem

Currently, there's an inconsistency between template definitions and module documentation:
- **Templates**: 78% (21 of 27) use `"main"` as the agent name
- **Module documentation**: Most of examples use `"example"` as a placeholder, which doesn't exist in any templates, except some modules who already use "main"

This creates friction when users copy examples from module docs into their templates that use `main` as the agent name.

## Solution

Update all module documentation to use `coder_agent.main` since:
1. It matches the most common convention (21 of 27 existing templates)
2. Creates consistency between docs and real-world usage
3. Reduces the need for users to edit copy-pasted examples

### Module Documentation (52 files)
- ✅ Updated 50+ module README.md files from `coder_agent.example` → `coder_agent.main`
- ✅ Updated root `README.md`
- ✅ Updated `CONTRIBUTING.md`
- ✅ Updated `examples/templates/main.tf`

### What's NOT in this PR

❌ Template changes - Templates are left as-is to avoid breaking existing workspaces  
❌ Linting rules - Will be addressed separately if desired

## Impact

✅ **Zero breaking changes**: Only documentation is updated  
✅ **Better DX**: Copy-paste examples work with 78% of existing templates  
✅ **Consistency**: Documentation now matches the de facto standard

## Testing

- ✅ All changes verified with `git diff`
- ✅ Only modified documentation files (52 files changed)
- ✅ 200 insertions, 200 deletions (pure substitution)

## Related

This addresses the documentation inconsistency identified in the registry audit. Template standardization can be addressed in a future PR if desired.

PR by @phorcys420
Non-blink follow-up to #566

### DIFF
--- a/registry/coder/modules/amazon-dcv-windows/README.md
+++ b/registry/coder/modules/amazon-dcv-windows/README.md
@@ -19,7 +19,7 @@ module "dcv" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/amazon-dcv-windows/coder"
   version  = "1.1.1"
-  agent_id = resource.coder_agent.main.id
+  agent_id = coder_agent.main.id
 }
 
 resource "coder_metadata" "dcv" {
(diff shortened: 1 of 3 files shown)


DevelopmentCats (approved):

matifali:
Thanks @phorcys420, a good follow-up would be to update all example templates to use `main` as the name for the `coder_agent` resource.

PR #575, opened 2025-12-02
feat: add vault-cli module with optional token configuration

### LINKED ISSUE #50: vault-cli: create a new moudle that installs vault CLI in a workspace and puts it in the PATH
Currently, we install vault CLI as part of our other vault modules. Let us create a new module that only installs the vault CLI without the authentication steps.

It should be an easy lift as we have all the installation-related code in all vault modules.

Part of coder/registry#49

matifali:
Closes #50

This adds a new `vault-cli` module that:
- Installs the Vault CLI
- Optionally configures token authentication if provided
- Optionally configures Vault Enterprise namespace if provided
- Sets up `VAULT_ADDR` environment variable
- Conditionally sets `VAULT_TOKEN` environment variable when token is provided
- Conditionally sets `VAULT_NAMESPACE` environment variable when namespace is provided
- Validates `vault_cli_version` must be `latest` or a semantic version (x.y.z)

The module can be used standalone for just CLI installation, or with a token and/or namespace for authentication scenarios.

## Follow-up

As a follow-up, we will update the other vault modules (`vault-github`, `vault-jwt`, `vault-token`) to use this module for CLI installation, reducing code duplication.


registry/coder/modules/vault-cli/run.sh
```diff
@@ -0,0 +1,116 @@
...
+    printf "Unsupported architecture: $${ARCH}\n"
+    return 1
+  fi
+
+  # Determine OS
+  OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+  # Fetch the latest version of Vault if VAULT_CLI_VERSION is 'latest'
+  if [ "$${VAULT_CLI_VERSION}" = "latest" ]; then
+    LATEST_VERSION=$(curl -s https://releases.hashicorp.com/vault/ | grep -v 'rc' | grep -oE 'vault/[0-9]+\.[0-9]+\.[0-9]+' | sed 's/vault\///' | sort -V | tail -n 1)
```
rowansmithau: As part of this overhaul we should move away from these grep and sed hacky approaches and use the official releases API - https://releases.hashicorp.com/docs/api/v1/

rowansmithau:
given run.sh has reqs like jq, unzip, wget/curl etc, maybe this should be noted in README.md so consumers know what is needed to use this upfront?

PR #587, opened 2025-12-09
fix(positron): fix readme version and verification status

matifali:
fixes the inconsistent version in the README and corrects the verification status


matifali:
@DevelopmentCats, please look into why the version check is flaking.

matifali:
@35C4n0r Thank you. Doing it in #588

PR #591, opened 2025-12-09
fix(mux): delete server.lock before starting server

ibetitsmike:
## Summary

Remove stale `~/.mux/server.lock` file before starting the mux server to prevent startup failures when a previous server didn't clean up properly.

## Changes

- Added `rm -f "$HOME/.mux/server.lock"` at the start of the `run_mux` function
- Bumped version to 1.0.5

## Testing

- All Terraform tests pass
- Shellcheck passes with only style warnings (unrelated to this change)


matifali (approved):
Assuming you tested. I am fine. LGTM. 🚢

PR #604, opened 2025-12-13
fix(zed): fix settings JSON parsing with base64 encoding

matifali:
## Problem

The Zed module's settings parsing was broken. The previous implementation used quote escaping:

```hcl
SETTINGS_JSON='${replace(var.settings, "\"", "\\\"")}'
```

This produced invalid JSON like `{\"theme\":\"dark\"}` which **jq could not parse** because the backslash-escaped quotes are not valid JSON syntax.

## Solution

Changed to use base64 encoding internally:

```hcl
locals {
  settings_b64 = var.settings != "" ? base64encode(var.settings) : ""
}

# In the script:
SETTINGS_B64='${local.settings_b64}'
SETTINGS_JSON="$(echo -n "${SETTINGS_B64}" | base64 -d)"
```

**User interface remains the same** - users still provide plain JSON via `jsonencode()` or heredoc:

```hcl
module "zed" {
  source   = "..."
  agent_id = coder_agent.main.id
  settings = jsonencode({
    theme    = "dark"
    fontSize = 14
  })
}
```

## Testing

Added comprehensive tests:

**Terraform tests (5):**
- URL generation (default, folder, agent_name)
- Settings base64 encoding verification
- Empty settings edge case

**Container e2e tests (3):**
- Creates settings file with correct JSON (including special chars: quotes, backslashes, URLs)
- Merges with existing settings via jq
- Exits early with empty settings

Also fixed existing tests to use `override_data` for proper workspace mocking.


DevelopmentCats:
Ill load this up and verify in the meantime

DevelopmentCats:
> > @matifali Have you tested this by chance? When I run this it looks like the project settings are read from `/home/coder/.zed/settings.json` rather than `/home/coder/.config/zed/settings.json`
> 
> This could be an upstream bug: As per the [docs](https://zed.dev/docs/remote-development#zed-settings), it should read from `~/.config/zed/settings.json` on Linux. :/

@matifali The only reason that I question this, is that I used the button to open my ubuntu workspace in zed, and when I click 

zed > Open Project Settings file

 it opens directly to `.zed/ settings.json`
 
 and the settings from this file that I wrote into it are not being applied.

Issue #609, opened 2025-12-16
fix: add pre_start_script support to claude-code module for dependency ordering


jasonwbarnett:
## Problem

When using multiple startup modules in a Coder workspace template (e.g., git-clone and claude-code), there's a race condition where both modules' startup scripts execute in parallel. This causes issues when one module depends on another.

### Specific Case: git-clone → claude-code

The claude-code module initializes with a workdir that should be populated by the git-clone module. However:

1. All workspace agent startup scripts execute **in parallel** via Go's errgroup.Group
2. The git-clone module's post_clone_script (which creates the workdir) runs sequentially AFTER the clone completes
3. The claude-code module's startup scripts run concurrently and may fail if they try to access the workdir before git-clone completes

### Root Cause

The Coder agent executes all startup scripts with  concurrently. There's no built-in mechanism to enforce sequential execution between different coder_script resources or modules.

## Solution

Add a  variable to the claude-code module that allows users to run custom shell commands **before** the Claude Code startup script executes. This enables dependency management:

1. claude-code's pre_start_script can wait for markers/conditions from other modules
2. Once preconditions are met, the main start script proceeds
3. This provides a clean way to handle module dependencies without architectural changes

## Use Case

This pattern is already used in other parts of the Coder registry (pre_install_script, post_install_script) and would provide consistency.

PR #613, opened 2025-12-16
docs(claude-code): document pre_install_script for module dependency ordering

jasonwbarnett:
## Summary

Clarifies that the existing `pre_install_script` variable can be used to handle dependencies between modules during workspace startup.

## Problem

When using multiple startup modules (e.g., git-clone and claude-code), there's a race condition where scripts execute in parallel. Module dependencies need to be managed, such as ensuring git-clone completes before Claude Code tries to access a workdir.

## Solution

The existing `pre_install_script` variable already provides this capability. Updated documentation to clarify this use case.

## Example

```hcl
module "claude-code" {
  source = "registry.coder.com/coder/claude-code/coder"
  
  workdir = "/path/to/repo"
  
  # Wait for git-clone to complete before starting
  pre_install_script = <<-EOT
    #!/bin/bash
    set -e
    while [ ! -f /tmp/.git-clone-complete ]; do
      sleep 1
    done
  EOT
}
```

Resolves issue #609.


DevelopmentCats (commented):
LGTM but we will need to update the version here for the change to reflect

PR #639, opened 2026-01-08
feat(cmd/tests): update tests to fail on invalid and absolute urls

35C4n0r:
## Description
Updated the tests:
- We now don't accept absolute urls (http / https)
- We only accept urls that start with `../../../../.icons/*` and exists

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally


cmd/readmevalidation/testSamples/modules/fail/wrongPathFormat.md
```diff
@@ -0,0 +1,19 @@
+---
+display_name: "Wrong Path"
+description: "Test module with wrong icon path format"
+icon: "../../../../.icons/invalid.svg"
```
matifali: We have the same repeated path in all modules and templates with the prefix `../../../../.icons/`. It would be nice if we could infer it and simplify the metadata format.

This is a topic for a follow-up PR, though.

35C4n0r: Oh, that's a good idea.

> This is a topic for a follow-up PR, though.
👍🏻

PR #652, opened 2026-01-12
fix(coder/modules/claude-code): update default boundary version and add tests.

35C4n0r:
## Description
Update boundary version to default to 'latest' and adjust related tests

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v4.3.1`  
**Breaking change:** [ ] Yes [ ] No

## Testing & Validation

- [ ] Tests pass (`bun test`)
- [ ] Changes tested locally

### DIFF
--- a/registry/coder/modules/claude-code/main.test.ts
+++ b/registry/coder/modules/claude-code/main.test.ts
@@ -87,7 +87,7 @@ describe("claude-code", async () => {
         enable_boundary: "true",
       },
     });
-    await execModuleScript(id, coderEnvVars)
+    await execModuleScript(id, coderEnvVars);
 
     await expectAgentAPIStarted(id);
 


35C4n0r:
Tests are bound to fail and are dependent on https://codercom.slack.com/archives/C094GRQG7UZ/p1768189738799269

registry/coder/modules/claude-code/main.tf
```diff
@@ -206,8 +206,8 @@ variable "enable_boundary" {
 
 variable "boundary_version" {
   type        = string
-  description = "Boundary version, valid git reference should be provided (tag, commit, branch)"
-  default     = "main"
+  description = "Boundary version, defaults to latest."
+  default     = "latest"
```
matifali: I think we should still support `main` as it builds from `main` branch instead of using the latest release. 
Good to default to `latest` to always get the latest published tagged release.

PR #662, opened 2026-01-16
feat: add zizmor workflow checks for GitHub Actions

CommanderK5:
## Description

This PR adds a zizmor security scan to the CI pipeline to analyze new and existing GitHub Actions workflows under .github/workflows/.

- Runs zizmor on PRs and fails the check when HIGH severity (or above) issues are found, so they can block merges.
- Runs zizmor on main to produce security reporting (where applicable), keeping visibility into findings over time.
- Intended to be added as a required status check so workflow-security regressions can’t land unnoticed.

Reference: coder/registry#642 (comment) / zizmor-action
## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [x] Other - CI / security tooling

## Testing & Validation
- Validation via PR check - opened a test PR with a deliberately risky workflow and confirmed zizmor reports and blocks on HIGH findings

## Related Issues
coder/registry#642 (comment) / zizmor-action

### DIFF
--- a/.github/workflows/check_registry_site_health.yaml
+++ b/.github/workflows/check_registry_site_health.yaml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - name: Run check.sh
         run: |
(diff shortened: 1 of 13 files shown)


jdomeracki-coder:
Also I think that we should review the blocking findings from:
https://github.com/coder/registry/actions/runs/21075932061/job/60617647791?pr=662

Switching to checksums instead of tags is for the most part a one time change

PR #686, opened 2026-01-30
Don't create CLAUDE_API_KEY coder_env if not set

AndreasSko:
## Description

At the moment, the `CLAUDE_API_KEY` coder_env will always be created, even if the variable itself is not. This can lead to the environment variable being unset if it has been set outside of Terraform.

With this PR, we make the `claude_api_key` coder_env conditional, so it will only be created if an API key has been set.

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/claude-code/main.tf`  
**New version:** `v4.7.4`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

None

### DIFF
--- a/registry/coder/modules/claude-code/README.md
+++ b/registry/coder/modules/claude-code/README.md
@@ -13,7 +13,7 @@ Run the [Claude Code](https://docs.anthropic.com/en/docs/agents-and-tools/claude
 ```tf
 module "claude-code" {
   source         = "registry.coder.com/coder/claude-code/coder"
-  version        = "4.7.3"
+  version        = "4.7.4"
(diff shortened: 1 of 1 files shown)


35C4n0r (changes requested):
@AndreasSko, LGTM, can you pull and run `./.github/scripts/version-bump.sh patch`.

AndreasSko:
@35C4n0r  done 👍

PR #737, opened 2026-02-20
feat(coder/modules/git-config): set user name and email via git config

nrvnrvn:
## Description

The likes of JetBrains IDEs rely on git config rather than env variables and prompt users upon first commit/push attempts.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/git-config`  
**New version:** `v1.0.34`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

None

### DIFF
--- a/registry/coder/modules/git-config/main.tf
+++ b/registry/coder/modules/git-config/main.tf
@@ -63,28 +63,59 @@ data "coder_parameter" "username" {
   })
 }
 
+locals {
+  git_user_name  = coalesce(try(data.coder_parameter.username[0].value, ""), data.coder_workspace_owner.me.full_name, data.coder_workspace_owner.me.name)
+  git_user_email = coalesce(try(data.coder_parameter.user_email[0].value, ""), data.coder_workspace_owner.me.email)
+}
+
 resource "coder_env" "git_author_name" {
   agent_id = var.agent_id
   name     = "GIT_AUTHOR_NAME"
-  value    = coalesce(try(data.coder_parameter.username[0].value, ""), data.coder_workspace_owner.me.full_name, data.coder_workspace_owner.me.name)
+  value    = local.git_user_name
 }
 
 resource "coder_env" "git_commmiter_name" {
   agent_id = var.agent_id
   name     = "GIT_COMMITTER_NAME"
-  value    = coalesce(try(data.coder_parameter.username[0].value, ""), data.coder_workspace_owner.me.full_name, data.coder_workspace_owner.me.name)
+  value    = local.git_user_name
 }
 
 resource "coder_env" "git_author_email" {
   agent_id = var.agent_id
   name     = "GIT_AUTHOR_EMAIL"
-  value    = coalesce(try(data.coder_parameter.user_email[0].value, ""), data.coder_workspace_owner.me.email)
+  value    = local.git_user_email
   count    = data.coder_workspace_owner.me.email != "" ? 1 : 0
 }
 
 resource "coder_env" "git_commmiter_email" {
   agent_id = var.agent_id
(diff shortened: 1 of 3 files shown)


registry/coder/modules/git-config/main.tf
```diff
@@ -63,28 +63,60 @@ data "coder_parameter" "username" {
...
+  display_name = "Configure git user email globally"
+  script       = <<EOT
+    #!/bin/bash
+    set -o errexit
+    set -o pipefail
+
+    git config --global user.email "${local.git_user_email}"
+  EOT
+  count        = data.coder_workspace_owner.me.email != "" ? 1 : 0
+}
```
matifali: Can we not combine them into a single script?

matifali (dismissed):

registry/coder/modules/git-config/main.tf
DevelopmentCats: I would agree here this definitely could be something we do with a `coder_script` named `git_user_config` or something to that effect since they are rather short, and identical otherwise.

nrvnrvn: The only reason for split here is to mirror the relevant `coder_env` logic where committer and author emails are set depending on the count ternary condition. I wonder why it is done that way in the first place specifically for email and not for name

matifali: IIRC, the only reason why I  did it separately is that with `coder_env`, we can only add 1 at a time.

DevelopmentCats: Pushed [f797a85](https://github.com/coder/registry/commit/f797a8503d86f7519708f24c0f58812b9a83f9b2) to address this. Both scripts are now collapsed into a single `coder_script.git_user_config` that always sets `user.name` and conditionally renders the `user.email` line using a `%{ if data.coder_workspace_owner.me.email != "" }` template directive (preserving the existing gate). `local.git_user_email` is wrapped in `try()` so it safely returns `""` when both the parameter and the workspace owner email are empty, since the combined script always references that local. Tests pass locally (7/7).

<sub>Generated by Coder Agents on behalf of @DevelopmentCats.</sub>

DevelopmentCats:
Pushed [`f797a85`](https://github.com/coder/registry/commit/f797a8503d86f7519708f24c0f58812b9a83f9b2) to address @matifali's review.

**Change:** collapsed `coder_script.git_config_user_name` and `coder_script.git_config_user_email` into a single `coder_script.git_user_config`. `user.name` is always set; the `user.email` line is rendered only when `data.coder_workspace_owner.me.email != ""`, matching the existing `coder_env` gate. `local.git_user_email` is wrapped in `try()` so the local safely returns `""` when both inputs are empty (the combined script always references it, even when no email line is rendered).

**Tests:** all 7 existing test cases pass locally. Each scenario expects one fewer resource (`-1` `coder_script`).

@matifali could you take another look when you have a moment? @nrvnrvn, feel free to revert any of this if you'd rather take a different path.

<sub>Generated by Coder Agents on behalf of @DevelopmentCats.</sub>

nrvnrvn:
@DevelopmentCats thanks for picking this up. That's fine, the only thing that matters is the end result.

DevelopmentCats:
> @DevelopmentCats thanks for picking this up. That's fine, the only thing that matters is the end result.

Sorry for the random bump haha. I am just doing some spring cleaning and trying to close/merge what we can before we implement some new contribution guidelines later on down the line 😄

DevelopmentCats (approved):

PR #769, opened 2026-02-27
feat(agentapi): add agentapi_cache_dir for persistent binary caching

iamriajul:
## Description

Add an opt-in `agentapi_cache_dir` variable to avoid re-downloading the AgentAPI binary on every workspace start. When set, the binary is stored in that directory after the first download and reused on subsequent starts — useful with a shared persistent volume.

The binary is cached with a versioned filename (e.g. `agentapi-linux-amd64-v0.10.0`) so multiple versions coexist safely. When `agentapi_version = "latest"`, the actual release tag is resolved via GitHub's redirect before computing the cache key, so the cache never gets stuck on a stale binary. The resolution request is skipped entirely when `agentapi_cache_dir` is not set.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/agentapi`
**New version:** `v2.2.0`

## Testing & Validation

- [x] Tests pass (`bun test` — all agentapi-specific tests pass)
- [x] Changes tested locally

## Related Issues

None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -149,3 +149,4 @@ gha-creds-*.json
 
 # IDEs
 .idea
+.serena
--- a/registry/coder/modules/agentapi/README.md
+++ b/registry/coder/modules/agentapi/README.md
@@ -16,7 +16,7 @@ The AgentAPI module is a building block for modules that need to run an AgentAPI
 ```tf
 module "agentapi" {
   source  = "registry.coder.com/coder/agentapi/coder"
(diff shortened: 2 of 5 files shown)


DevelopmentCats:
Agentapi is being deprecated we are going to close this PR unless Jay wants to reopen it for any reason

PR #771, opened 2026-02-27
fix(registry/coder/modules): write bootstrap scripts to module dir instead of /tmp

mafredri:
Bootstrap scripts used fixed /tmp paths with generic names like
/tmp/main.sh, making them easy to collide with or accidentally
overwrite. Write them into $HOME/<module_dir_name>/scripts/ instead,
following the pattern used by the agent-helper module.

Also fix install.sh hardcoding port 3284 instead of using
ARG_AGENTAPI_PORT.

Refs #736


DevelopmentCats:
closing because of agentapi no longer being supported within the agent modules

PR #798, opened 2026-03-12
fix(code-server,vscode-web): parse JSONC in extensions.json and .code-workspace

### LINKED ISSUE #708: auto_install_extensions fails on JSONC comments and trailing commas
**Module:** `code-server` (1.4.3) and `vscode-web` (1.5.0)

**Expected:** `auto_install_extensions = true` parses `.vscode/extensions.json` with comments and trailing commas (valid JSONC).

**Actual:** The `sed 's|//.*||g' | jq` pipeline fails on:
- `/* */` block comments — `sed` only strips `//`
- Trailing commas — `jq` rejects them

> **Note:** `sed` also mangles `//` inside strings, but `recommendations[]` only accepts extension IDs matching vsce's [`nameRegex`](https://github.com/microsoft/vscode-vsce/blob/82cd05b077dfda69c80cc15b3bc7a266965f7880/src/validation.ts#L5) (`[a-z0-9\-]`), so this is not a practical concern.

### Why this is valid JSONC

VS Code itself parses `extensions.json` with its built-in JSONC parser ([`src/vs/base/common/json.ts`](https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/base/common/json.ts)) using [default options](https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/base/common/json.ts#L138-L140) that allow line comments (`//`), block comments (`/* */`), and trailing commas. The file that reads workspace extension recommendations ([`workspaceExtensionsConfig.ts`](https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/workbench/services/extensionRecommendations/common/workspaceExtensionsConfig.ts)) calls `parse()` with these defaults, so all three features are officially supported in `.vscode/extensions.json`.

### Affected code

The same `sed | jq` pattern appears in three places:

**code-server:**
https://github.com/coder/registry/blob/f1748c80f7d123d9f3a7867ea1835feaf227d1f2/registry/coder/modules/code-server/run.sh#L133

**vscode-web** (workspace file):
https://github.com/coder/registry/blob/f1748c80f7d123d9f3a7867ea1835feaf227d1f2/registry/coder/modules/vscode-web/run.sh#L161

**vscode-web** (folder fallback):
https://github.com/coder/registry/blob/f1748c80f7d123d9f3a7867ea1835feaf227d1f2/registry/coder/modules/vscode-web/run.sh#L173

### Reproduction

Create `.vscode/extensions.json`:
```jsonc
{
  "recommendations": [
    "ms-python.python",
    /* a block comment */
    "dbaeumer.vscode-eslint", // trailing comma
  ]
}
```
Use `auto_install_extensions = true` — extensions fail to install.

### Error output

Block comments cause `jq` to fail (the `/* ... */` syntax passes through `sed` untouched):

```
jq: parse error: Invalid numeric literal at line 4, column 7
```

When `sed` strips a line comment but leaves the trailing comma, `jq` also fails:

```
jq: parse error: Expected another array element at line 5, column 3
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rodmk:
Fixes coder/registry#708. Supersedes [#707](<https://github.com/coder/registry/issues/707>).

## Description

`auto_install_extensions` fails to parse `.vscode/extensions.json` and `.code-workspace` files that contain valid JSONC (block comments, line comments, trailing commas). VS Code officially supports these via its built-in JSONC parser with [default options](<https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/base/common/json.ts#L138-L140>).

This PR replaces the original `sed 's|//.*||g' | jq` pipeline with a small, portable `strip_jsonc_for_extensions` helper (defined once per script) that strips JSONC features before handing the file to `jq` — no new dependencies, no GNU-only extensions.

## Type of Change

- [X] Bug fix

## Module Information

**Path:** `registry/coder/modules/code-server` and `registry/coder/modules/vscode-web`<br>**New version:** code-server `1.5.2`, vscode-web `1.6.2`<br>**Breaking change:** No

## Problem

The original `sed 's|//.*||g' | jq` pipeline only handled line comments; block comments and trailing commas broke `jq`. Naive fixes introduce two portability traps and one ordering bug:

1. `sed -z` (slurp the whole file) is a GNU extension — it does not exist on BSD sed (macOS) and would break there.
2. `[^\n]` **inside a bracket expression** is a GNU extension — BSD sed reads `\n` in a bracket as a literal backslash and `n`, so `//[^\n]*` silently corrupts any extension ID containing the letter `n` (e.g. `dbaeumer.vscode-eslint`).
3. **Stripping line comments before block comments** corrupts a `//` that lives *inside* a block comment (e.g. a URL in `/* see https://open-vsx.org */`), leaving an unterminated `/*` that makes `jq` fail — installing zero extensions with no error surfaced. A single-line/minified file was also dropped entirely by the `:a;N;$!ba` slurp idiom.

## Fix

`strip_jsonc_for_extensions` runs three ordered `sed` passes, each scoped to what it needs and portable across GNU (Linux), BSD (macOS), and BusyBox (Alpine) sed:

```bash
strip_jsonc_for_extensions() {
  sed -E ':a
$!{
N
ba
}
s#/[*]([^*]|[*]+[^*/])*[*]+/##g' "$1" \
    | sed -E 's#^[[:space:]]*//.*##; s#([^:])//.*#\1#' \
    | sed -E ':a
$!{
N
ba
}
s/,[^]}"]*([]}])/\1/g'
}
```

1. **Block comments** first — slurps the whole file so `/* ... */` can span lines. Running first means a URL inside a block comment is removed as a unit before the line-comment pass ever sees its `//`.
2. **Line comments** per line — `// ...` stops at end of line without the non-portable `[^\n]` class. A `//` preceded by `:` is preserved, so URLs in string values (e.g. `"http.proxy": "https://…"` in a `.code-workspace`) survive.
3. **Trailing commas** — slurps the whole file so a comma and its closing bracket may sit on different lines.

The `:a;$!{N;ba}` slurp idiom is single-line safe (it falls through on the last/only line), unlike `:a;N;$!ba`, which drops single-line input.

Also makes the **code-server** recommendations query null-safe to match vscode-web (`.recommendations[]` → `(.recommendations // [])[]`), so a valid `extensions.json` with no `recommendations` key no longer errors.

### Known limitation

This is not a general-purpose JSONC parser. A **non-URL** `//` inside a string value (e.g. a regex setting `"a//b"`) can still be mis-stripped — only a real string-aware parser could distinguish it, and a hand-rolled parser was [rejected in #707](<https://github.com/coder/registry/pull/707#issuecomment-3967390449>) as over-engineered. Extension IDs are restricted to `[a-z0-9\-.]` by vsce's [`nameRegex`](<https://github.com/microsoft/vscode-vsce/blob/82cd05b077dfda69c80cc15b3bc7a266965f7880/src/validation.ts#L5>), and URLs (the common `//`-in-string case) are handled by the `://` guard, so the recommendations arrays this feature actually reads are safe.

### Alternatives considered

| Approach | Tradeoff |
| -- | -- |
| `sed -z` | Simpler slurp, but `-z` is a GNU extension — breaks on macOS BSD sed |
| **Single** `sed -E` **with** `//[^\n]*` | Works on GNU sed, but `[^\n]` in a bracket is not portable — BSD sed treats it as literal `\` + `n`, silently corrupting IDs containing `n` |
| `perl -0777 -pe` | Clean one-liner, but no module in the repo uses perl and [Apple may remove it from future macOS](<https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes>) |
| **Node.js state machine** | Correct JSONC parsing, but [rejected in #707](<https://github.com/coder/registry/pull/707#issuecomment-3967390449>) as over-engineered |
| **Python3** | Could parse JSONC natively, but not guaranteed on all workspaces and would need a `command -v` guard |
| **Three-pass** `sed` (this PR) | Two extra pipes, but zero new dependencies, fully portable across GNU/BSD/BusyBox sed, and fixes comment/URL ordering + single-line + null-safety |

## Testing & Validation

- [X] `terraform test`: code-server 8/8, vscode-web 3/3
- [X] code-server + vscode-web TypeScript suites pass in CI (incl. the container `installs and runs code-server` test)
- [X] Container tests cover the JSONC extension path and the null-safe `recommendations` guard in **both** modules: a valid `extensions.json` with no `recommendations` key installs nothing and produces no `jq` error, and the tests assert the auto-install path actually ran
- [X] `prettier` (with `prettier-plugin-sh`) and `shellcheck --severity=warning` clean; `terraform fmt` clean
- [X] `templatefile()` renders both `run.sh` files; rendered scripts pass `bash -n`
- [X] Version bumped

The `strip_jsonc_for_extensions` helper was validated with a 25-case matrix (the original 19 JSONC cases plus new regression cases for URL-in-block-comment, single-line/minified input, IDs containing `n`, a line comment containing a quote before a trailing comma, and URL-valued `.code-workspace` settings). The matrix was run against the exact bytes committed to both `run.sh` files and **passes 25/25 on GNU sed (Ubuntu), BSD sed (macOS), and BusyBox sed (Alpine)**.

## Related Issues

* Fixes coder/registry#708
* Supersedes [#707](<https://github.com/coder/registry/issues/707>)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>) using Claude Opus 4.8 and Claude Fable 5


registry/coder/modules/vscode-web/run.sh
```diff
@@ -150,15 +150,26 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
 
+# Strip JSONC features (block/line comments, trailing commas) for jq parsing
+strip_jsonc_for_extensions() {
+  sed 's|//.*||g' "$1" | sed -E '
+    :a
+    N
+    $!ba
+    s#/[*]([^*]|[*]+[^*/])*[*]+/##g
+    s/,[^]}"]*([]}])/\1/g
+  '
```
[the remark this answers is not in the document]
rodmk: @DevelopmentCats Would you prefer this approach? https://github.com/coder/registry/pull/707 was rejected explicitly due to being over-engineered (I don't disagree - hand rolling a full jsonc stripping state machine is a lot) so I'm getting mixed signals.

DevelopmentCats (approved):
I agree with you that the Copilot Comments were wrong.

This LGTM though thanks for the fix and quick response!! 😸

35C4n0r:
@rodmk can you resolve the conflicts.

rodmk:
@35C4n0r Whoa, I didn't realize this hadn't been merged yet, it was approved months ago 😅 I'll resolve the conflicts promptly.

rodmk:
Resolved the conflicts against the latest `main`. The only real overlap was in the `code-server`/`vscode-web` `run.sh` scripts: I merged the new `WORKSPACE`-file support together with this PR's JSONC parsing, so both commented `extensions.json` files and `.code-workspace` files work. Version bumps re-based onto the current releases — `code-server` → `1.5.1`, `vscode-web` → `1.6.1`.

Validation: `fmt`, README validation, golangci-lint, and the full `code-server` + `vscode-web` test suites all pass. (The one failing test is an unrelated `nexus-repository` timeout flake.)

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.8

registry/coder/modules/code-server/run.sh
```diff
@@ -116,6 +116,17 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
+# Strip JSONC features (block/line comments, trailing commas) for jq parsing
+strip_jsonc_for_extensions() {
+  sed 's|//.*||g' "$1" | sed -E '
+    :a
+    N
+    $!ba
+    s#/[*]([^*]|[*]+[^*/])*[*]+/##g
+    s/,[^]}"]*([]}])/\1/g
+  '
+}
```
35C4n0r: **P1 — `sed 's|//.*||g'` runs before block-comment removal; two silent failures result.**

1. **Inline block comment containing `://`** — a line like `/* https://open-vsx.org */` has `//open-vsx.org */` stripped by stage 1, leaving `/* https:` in the buffer. Stage 2 finds no closing `*/` and passes the fragment through. `jq` then fails with `parse error: Invalid numeric literal` — zero extensions installed, nothing printed to the user. Verified locally.

2. **URL-valued settings in `.code-workspace` files** — `"http.proxy": "https://proxy.corp:8080"` becomes `"http.proxy": "https:` (unclosed string literal). `jq` exits with `Invalid string: control characters from U+0000 through U+001F must be escaped`. Same silent failure. Verified locally.

The suggestion merges into a single `sed -E` pass with block comments removed first. `s#//[^\n]*##g` then handles line comments on the slurped buffer — `[^\n]*` stops at each embedded newline, so it is equivalent to per-line stripping without a separate process. This fixes failure #1. Failure #2 (URL string values) is a fundamental limit of the regex approach — only a JSONC-aware parser can reliably distinguish `//` inside a string from `//` in a comment.

```suggestion
# Strip JSONC features (block/line comments, trailing commas) for jq parsing
strip_jsonc_for_extensions() {
  sed -E '
    :a
    N
    $!ba
    s#/[*]([^*]|[*]+[^*/])*[*]+/##g
    s#//[^\n]*##g
    s/,[^]}"]*([]}])/\1/g
  ' "$1"
}
```

> *Generated by Coder Agents on behalf of @35C4n0r.*

registry/coder/modules/vscode-web/run.sh
```diff
@@ -150,15 +150,26 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
+}
```
35C4n0r: **P1 — URL corruption / single-line file: stage ordering is wrong**

The current implementation pipes through a line-comment stripper *before* slurping into a multi-line buffer:

```sh
sed 's|//.*||g' "$1" | sed -E '
  :a
  N
  $!ba
  s#/[*]([^*]|[*]+[^*/])*[*]+/##g
  ...
```

Two verified failure modes:

1. **URL inside a block comment** — `/* https://open-vsx.org */`: stage 1 strips `://open-vsx.org */`, leaving `/* https:` unclosed. Stage 2's block-comment regex can't match, jq receives garbage and exits non-zero — zero extensions installed, no error surfaced.

2. **URL-valued setting** — `"http.proxy": "https://proxy.corp"`: stage 1 strips `://proxy.corp"`, producing an unclosed string literal. jq exits 5, zero extensions installed silently.

3. **Single-line input** — `:a;N;$!ba` exits immediately at EOF without running any substitution (GNU sed behaviour on last-line `N`). A one-line `{"recommendations":["ext1",]}` is passed through unchanged and jq fails on the trailing comma.

Fix: slurp the whole file first, then strip block comments, then line comments — in a single `sed` invocation:

```suggestion
strip_jsonc_for_extensions() {
  sed -E '
    :a
    N
    $!ba
    s#/[*]([^*]|[*]+[^*/])*[*]+/##g
    s#//[^\n]*##g
    s/,[^]}"]*([]}])/\1/g
  ' "$1"
}
```

`s#//[^\n]*##g` is safe on the slurped buffer because `[^\n]*` stops at each embedded newline, so it only strips to end-of-line.

> *Generated by Coder Agents on behalf of @35C4n0r.*

registry/coder/modules/code-server/run.sh
```diff
@@ -143,8 +154,8 @@ if [ "${AUTO_INSTALL_EXTENSIONS}" = true ]; then
 
   if [ -n "$RECOMMENDATIONS_FILE" ]; then
     printf "🧩 Installing extensions from %s...\n" "$RECOMMENDATIONS_FILE"
-    # Use sed to remove single-line comments before parsing with jq
-    extensions=$(sed 's|//.*||g' "$RECOMMENDATIONS_FILE" | jq -r "$RECOMMENDATIONS_QUERY")
+    # Strip JSONC comments and trailing commas before parsing with jq
+    extensions=$(strip_jsonc_for_extensions "$RECOMMENDATIONS_FILE" | jq -r "$RECOMMENDATIONS_QUERY")
```
35C4n0r, line 220: **P1 — `code-server` null-safety gap in `RECOMMENDATIONS_QUERY`**

This PR upgrades `vscode-web` to use a null-safe query:

```sh
# vscode-web/run.sh (this PR)
RECOMMENDATIONS_QUERY='(.recommendations // [])[]'
```

but `code-server` was not updated in parallel. The query at line 137 of `code-server/run.sh` is still:

```sh
RECOMMENDATIONS_QUERY=".recommendations[]"
```

A valid `extensions.json` with no `recommendations` key at all (permitted by the VS Code schema) causes `jq` to emit `null | .[]` which exits non-zero — silently discarding all extension installations on `code-server` while `vscode-web` handles the same file cleanly.

**Fix (line 137, not in the diff hunk above — apply manually):**

```diff
- RECOMMENDATIONS_QUERY=".recommendations[]"
+ RECOMMENDATIONS_QUERY='(.recommendations // [])[]'
```

Note the single-quote change: the expression contains no shell variables, so single quotes are correct and prevent accidental expansion.

> *Generated by Coder Agents on behalf of @35C4n0r.*

35C4n0r (changes requested):
Two P1 bugs found in the `strip_jsonc_for_extensions` shell function introduced by this PR. Both are verified with live tests. Each inline comment above includes a suggested fix or a plain diff.

**Summary of required changes:**

| # | File | Issue | Severity |
|---|------|-------|----------|
| 1 | `code-server/run.sh` lines 119–128 | Line-comment pass runs before block-comment slurp — corrupts URLs in comments and settings; single-line files pass through unprocessed | P1 |
| 2 | `vscode-web/run.sh` lines 153–162 | Same stage-ordering bug | P1 |
| 3 | `code-server/run.sh` line 137 | `RECOMMENDATIONS_QUERY` not null-safe (vscode-web was fixed, code-server was not) | P1 |

> *Generated by Coder Agents on behalf of @35C4n0r.*

rodmk:
Thanks for the careful review — all three findings were valid and are fixed in f29505a. I reproduced each one on BSD sed (macOS) first, which also turned up a portability issue in the suggested patch, so I took a slightly different route to fix the same bugs. Details below.

### 1 & 2 — line-comment pass running before the block-comment slurp (both files)

Confirmed, both failure modes reproduce:

- **URL inside a block comment** — `/* see https://open-vsx.org */` → the first-stage `s|//.*||g` strips `//open-vsx.org */`, leaving an unterminated `/* see https:` that `jq` rejects. Zero extensions, no error surfaced.
- **Single-line / minified file** — `:a;N;$!ba` drops single-line input entirely (on BSD it produced empty output).

**On the suggested fix:** collapsing into one `sed -E` with `s#//[^\n]*##g` fixes the block-comment case, but `[^\n]` inside a bracket expression is a GNU extension. On BSD sed (macOS, which this module supports via `platform = "darwin"`) `\n` in a bracket is read as a literal backslash + `n`, so `//[^\n]*` stops at the first `n` and silently corrupts IDs — e.g. `dbaeumer.vscode-eslint` came out as `nting extension`. Verified locally.

So I kept the fix portable with three ordered passes instead:

```bash
strip_jsonc_for_extensions() {
  sed -E ':a
$!{
N
ba
}
s#/[*]([^*]|[*]+[^*/])*[*]+/##g' "$1" \
    | sed -E 's#^[[:space:]]*//.*##; s#([^:])//.*#\1#' \
    | sed -E ':a
$!{
N
ba
}
s/,[^]}"]*([]}])/\1/g'
}
```

- **Block comments first** (whole-file slurp) so a `//` inside `/* ... */` is removed as a unit before the line-comment pass sees it.
- **Line comments per line** — no `[^\n]`; a `//` preceded by `:` is preserved so URLs in string values survive.
- **Trailing commas** last (slurp).
- `:a;$!{N;ba}` is single-line safe (falls through on the last/only line), fixing the minified-file case.

### 3 — code-server null-safety

Fixed: `RECOMMENDATIONS_QUERY` is now `(.recommendations // [])[]` (and the workspace query `(.extensions.recommendations // [])[]`), matching vscode-web.

### On failure mode 2 (URL in a `.code-workspace` setting value)

Agreed this is the fundamental limit you called out — a purely regex-based stripper can't reliably tell a `//` in a string from a comment. The `://`-guard in the line-comment pass handles the realistic case (URL-valued settings such as `"http.proxy": "https://proxy.corp:8080"` now survive). A **non-URL** `//` inside a string value is still out of reach without a real parser, which #707 established as over-engineered; I documented this as a known limitation in the PR description.

### Validation

The helper was run against a 25-case matrix (the original 19 JSONC cases + regressions for URL-in-block-comment, single-line input, IDs containing `n`, a quote inside a line comment before a trailing comma, and URL-valued `.code-workspace` settings), using the exact bytes committed to both `run.sh` files. **25/25 on GNU sed (Ubuntu), BSD sed (macOS), and BusyBox sed (Alpine).** `terraform test`, the code-server/vscode-web TS suites, `shellcheck`, and `prettier` all pass. Versions bumped to code-server `1.5.2` / vscode-web `1.6.1`.

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.8

35C4n0r:
@rodmk can you resolve the merge conflict.

rodmk:
Yup sorry just noticed, doing that now.

rodmk:
All set - no more conflicts for now.

35C4n0r (approved):

35C4n0r:
@rodmk can you add some tests in main.test.ts for both the modules to test the changes.

rodmk:
Added tests in `main.test.ts` for both modules (`b0fc629`) that drive the rendered script through the `auto_install_extensions` path with a mock CLI recording `--install-extension` calls — covering JSONC comments (incl. a URL inside a block comment), trailing commas, the null-safe query, and a `.code-workspace` with a URL-valued setting. They fail on the pre-fix code.

One thing worth flagging: `code-server` reaches the install path cleanly via `use_cached`, but `vscode-web` early-exits on `use_cached`/`offline`, so its test stubs `curl`/`tar` to no-op the download and let the pre-placed mock binary survive. All green in CI.



🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 4.8

.github/typos.toml
```diff
@@ -8,6 +8,7 @@ mavrickrishi = "mavrickrishi" # Username
 mavrick = "mavrick" # Username
 inh = "inh" # Option in setpriv command
 exportfs = "exportfs" # nfs related binary
+ba = "ba" # sed branch command in :a;N;$!ba pattern
```
35C4n0r:
```suggestion
ba = "ba" # sed branch command in :a;$!{N;ba} pattern
```

registry/coder/modules/code-server/run.sh
```diff
@@ -156,18 +156,51 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
+#      unit and its // is never seen by the line-comment pass.
+#   2. Line comments   - per line, so // ... stops at end of line without the
+#      non-portable [^\n] class (BSD sed reads \n inside a bracket as a literal
+#      backslash and n, silently corrupting IDs containing "n"). A // preceded
+#      by ':' is preserved so URLs in string values (e.g. proxy settings in a
+#      .code-workspace) survive.
+#   3. Trailing commas - slurps the whole file so a comma and its closing
+#      bracket may sit on different lines.
+# The ':a;$!{N;ba}' slurp is single-line safe (it falls through on the last or
+# only line), unlike ':a;N;$!ba', which drops single-line input entirely.
```
35C4n0r:
```suggestion
# only line).
```

registry/coder/modules/vscode-web/run.sh
```diff
@@ -150,15 +150,48 @@ for extension in "$${EXTENSIONLIST[@]}"; do
...
+#      unit and its // is never seen by the line-comment pass.
+#   2. Line comments   - per line, so // ... stops at end of line without the
+#      non-portable [^\n] class (BSD sed reads \n inside a bracket as a literal
+#      backslash and n, silently corrupting IDs containing "n"). A // preceded
+#      by ':' is preserved so URLs in string values (e.g. proxy settings in a
+#      .code-workspace) survive.
+#   3. Trailing commas - slurps the whole file so a comma and its closing
+#      bracket may sit on different lines.
+# The ':a;$!{N;ba}' slurp is single-line safe (it falls through on the last or
+# only line), unlike ':a;N;$!ba', which drops single-line input entirely.
```
35C4n0r:
```suggestion
# only line).
```

35C4n0r (changes requested):
Can you also add test for the null-safe guard on a missing recommendations key.

35C4n0r:
@rodmk, can you address the above reviews.

rodmk:
@35C4n0r Thanks for the review! I've addressed the feedback in [61d4956](https://github.com/coder/registry/pull/798/commits/61d4956b006f95152eebd262812ff05026554dcb). I've also had my agent self-review the changes, and I'll have it drop a note below in a sec.

rodmk:
Self-review note on [61d4956](https://github.com/coder/registry/pull/798/commits/61d4956b006f95152eebd262812ff05026554dcb), as promised above.

**TL;DR** — the requested null-safe guard test is added for vscode-web (code-server already had one), all three inline suggestions are applied, both tests were hardened against vacuous passes during self-review, and full CI is green (563 tests).

<details>
<summary><b>Review feedback addressed</b></summary>

- Added the vscode-web test for the null-safe guard: a valid `extensions.json` with no `recommendations` key must install nothing and produce no `jq` error ([vscode-web/main.test.ts#L388-L434](https://github.com/coder/registry/blob/61d4956b006f95152eebd262812ff05026554dcb/registry/coder/modules/vscode-web/main.test.ts#L388-L434)). code-server already had the equivalent test from the earlier test commit ([code-server/main.test.ts#L295-L340](https://github.com/coder/registry/blob/61d4956b006f95152eebd262812ff05026554dcb/registry/coder/modules/code-server/main.test.ts#L295-L340)).
- All three inline suggestions applied as suggested: the `typos.toml` comment now references the `:a;$!{N;ba}` pattern actually in use, and both `run.sh` comments drop the stale reference to the old slurp idiom.

</details>

<details>
<summary><b>Self-review findings applied (test hardening)</b></summary>

- Both null-safety tests now assert the auto-install path actually ran (`Installing extensions from` in stdout), so they cannot pass vacuously — previously, a failed `jq` install inside the test container would have green-lit the test without ever exercising the query.
- Swapped the exact `Cannot iterate over null` match for jq's stable `jq: error` prefix, so the assertion survives jq wording changes and catches any jq failure, not just null iteration.
- Mutation-checked: reverting the query to `.recommendations[]` makes the new test fail; with `(.recommendations // [])[]` it passes.

</details>

<details>
<summary><b>Self-review findings not acted on</b></summary>

- The sed passes remain string-context-unaware, so a non-URL `//` or `/*` inside a string value (e.g. glob settings) can still be mis-stripped. This is the documented "Known limitation" in the PR description — a string-aware parser was already rejected as over-engineered in #707.

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Fable 5

35C4n0r (approved):

PR #815, opened 2026-03-26
fix(opencode): pass VERSION to bash instead of curl in install pipe

megheaiulian:
## Summary

- Fix version pinning bug in the OpenCode install script (`registry/coder-labs/modules/opencode/scripts/install.sh`, line 42)

**Bug:** The install command was:
```bash
VERSION=$ARG_OPENCODE_VERSION curl -fsSL https://opencode.ai/install | bash
```

`VERSION` was set as an environment variable prefix to `curl` (the left side of the pipe), so the `bash` process on the right side of the pipe never received it. In a shell pipeline, each command runs in its own subprocess, so env var prefixes only apply to the immediately following command. This caused the installer script to always install the latest version instead of the pinned version specified by the user.

**Fix:** Move `VERSION` to prefix `bash` instead of `curl`:
```bash
curl -fsSL https://opencode.ai/install | VERSION=$ARG_OPENCODE_VERSION bash
```

Now the `VERSION` variable is correctly available to the install script executed by `bash`.

## Test plan

- [x] Set `opencode_version` to a specific version (e.g., `0.1.0`) and verify that version is installed instead of latest
- [x] Set `opencode_version` to `latest` and verify the latest version is still installed (this code path is unchanged)
- [x] Verify `opencode --version` output matches the requested version after install

### DIFF
--- a/registry/coder-labs/modules/opencode/README.md
+++ b/registry/coder-labs/modules/opencode/README.md
@@ -13,7 +13,7 @@ Run [OpenCode](https://opencode.ai) AI coding assistant in your workspace for in
 ```tf
 module "opencode" {
   source   = "registry.coder.com/coder-labs/opencode/coder"
-  version = "0.1.2"
+  version  = "0.1.2"
   agent_id = coder_agent.main.id
   workdir  = "/home/coder/project"
 }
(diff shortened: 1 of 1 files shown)


matifali:
Thanks for the fix.

PR #833, opened 2026-04-04
fix(copilot): remove hardcoded model enum to allow any Copilot model

### LINKED ISSUE #832: github copilot module only supports specific models
Currently there is a check in the github copilot module that only allows setting the model to one of three - and these are rather outdated at this point. Since models are changing fast, it would seem logical to remove the hard check that a model must be in a enum. As well, the default i think should be sonnet 4.6 instead of sonnet 4.5.

Some example use case is thst the user should be able to set models gpt-5-mini, gpt-5.4, sonnet-4.6, opus-4.6, etc...

matifali:
The `copilot_model` variable was restricted to a hardcoded enum of three models (`claude-sonnet-4`, `claude-sonnet-4.5`, `gpt-5`). Models change fast and this validation was blocking users from using newer models.

## Changes

- Remove `validation` block from `copilot_model` variable in `main.tf`
- Update variable description to indicate any Copilot-supported model can be used
- Replace enum validation test with a test that verifies arbitrary model strings are accepted
- Bump module version to `0.4.1` in README examples

Closes #832

> 🤖 This PR was created with the help of Coder Agents, and needs a human review. 🧑‍💻

### DIFF
--- a/registry/coder-labs/modules/copilot/README.md
+++ b/registry/coder-labs/modules/copilot/README.md
@@ -13,7 +13,7 @@ Run [GitHub Copilot CLI](https://docs.github.com/copilot/concepts/agents/about-c
 ```tf
 module "copilot" {
   source   = "registry.coder.com/coder-labs/copilot/coder"
-  version  = "0.4.0"
+  version  = "0.4.1"
   agent_id = coder_agent.example.id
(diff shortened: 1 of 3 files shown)


35C4n0r (approved):
LGTM!

PR #836, opened 2026-04-07
feat(jetbrains): skip HTTP calls when ide_config is set

### LINKED ISSUE #835: JetBrains module makes HTTP calls even when `ide_config` pins build numbers
## Problem

The `data "http" "jetbrains_ide_versions"` resource fires for every selected IDE, even when the user has pinned versions via `ide_config`. There is no way to skip the HTTP call entirely, even when the user already provides all the build info.

## Expected behavior

When `ide_config` is provided, the module should skip all HTTP calls and use the pinned build numbers directly.

## Fix

When `ide_config` is set (non-null), `data.http.jetbrains_ide_versions` `for_each` is empty, so zero HTTP calls are made.

Related: [ECO-29](https://linear.app/codercom/issue/ECO-29/bug-data-http-jetbrains-ide-versions-failure-can-be-fatal-even-when-no)

matifali:
Fixes #835

## Problem

The `data "http"` resource always fires for every selected IDE, even when the user has pinned versions via `ide_config`. In air-gapped or caching scenarios, this causes:

- **30-second hangs** when `releases_base_link` is set to a dummy URL like `https://localhost`
- **Fatal errors** with `https://localhost:1` (connection refused)
- The documented "air-gapped fallback" via `try()` never actually worked — the `http` data source fails before `try()` can catch anything

## Fix

When `ide_config` is provided, the module now skips all HTTP calls and uses the pinned build numbers directly.

| Scenario | `ide_config` | HTTP calls | Build source | On API failure |
|---|---|---|---|---|
| User wants latest | `null` (default) | Yes | JetBrains API | Terraform error (fail loudly) |
| User pins versions | Set | **None** | `ide_config.build` | N/A |

### Changes

- `ide_config` default changed from a full map to `null`
- `name` and `icon` are now `optional(string)` in `ide_config` — falls back to built-in metadata
- `data.http.jetbrains_ide_versions` `for_each` is empty when `ide_config` is set
- Static `ide_metadata` local provides name/icon when `ide_config` is null
- Removed `try()` fallback from `parsed_responses` — API errors are now explicit instead of silently using stale builds
- Cross-variable validation rejects `major_version`, `channel`, and `releases_base_link` when `ide_config` is set
- Validation for `ide_config ⊇ default` added (previously only `ide_config ⊇ options` was checked)
- Version bumped `1.3.1` → `1.4.0`

### Usage

```tf
module "jetbrains" {
  source   = "registry.coder.com/coder/jetbrains/coder"
  version  = "1.4.0"
  agent_id = coder_agent.main.id
  folder   = "/home/coder/project"

  # Zero HTTP calls — only build is required.
  ide_config = {
    "GO" = { build = "261.22158.291" }
    "PY" = { build = "261.22158.340" }
  }
  options = ["GO", "PY"]
}
```

> 🤖 This PR was created with the help of Coder Agents, and needs a human review. 🧑‍💻

### DIFF
--- a/registry/coder/modules/jetbrains/README.md
+++ b/registry/coder/modules/jetbrains/README.md
@@ -14,7 +14,7 @@ This module adds JetBrains IDE buttons to launch IDEs directly from the dashboar
 module "jetbrains" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
   agent_id = coder_agent.main.id
   folder   = "/home/coder/project"
 }
@@ -39,7 +39,7 @@ When `default` contains IDE codes, those IDEs are created directly without user 
 module "jetbrains" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
   agent_id = coder_agent.main.id
   folder   = "/home/coder/project"
   default  = ["PY", "IU"] # Pre-configure PyCharm and IntelliJ IDEA
@@ -52,7 +52,7 @@ module "jetbrains" {
 module "jetbrains" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
   agent_id = coder_agent.main.id
   folder   = "/home/coder/project"
   # Show parameter with limited options
@@ -66,7 +66,7 @@ module "jetbrains" {
 module "jetbrains" {
   count         = data.coder_workspace.me.start_count
   source        = "registry.coder.com/coder/jetbrains/coder"
-  version       = "1.3.1"
+  version       = "1.4.0"
   agent_id      = coder_agent.main.id
   folder        = "/home/coder/project"
   default       = ["IU", "PY"]
@@ -75,30 +75,37 @@ module "jetbrains" {
 }
 ```
 
-### Custom IDE Configuration
+### Pinned Versions (Air-Gapped / Cached)
+
+When `ide_config` is set, the module makes zero HTTP calls and uses the
+provided build numbers directly. This is ideal for air-gapped environments
+or when caching IDE installations.
+
+> [!TIP]
+> To find the latest build number for an IDE, query the JetBrains releases API:
+>
+> ```sh
+> curl -s "https://data.services.jetbrains.com/products/releases?code=GO&type=release&latest=true" | jq 'to_entries[0].value[0] | {build, version}'
+> ```
+>
+> Replace `GO` with the product code for the IDE you want (e.g. `IU`, `PY`, `CL`).
 
 ```tf
 module "jetbrains" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
   agent_id = coder_agent.main.id
-  folder   = "/workspace/project"
+  folder   = "/home/coder/project"
 
-  # Custom IDE metadata (display names and icons)
+  # Only build is required. Name and icon fall back to built-in defaults.
   ide_config = {
-    "IU" = {
-      name  = "IntelliJ IDEA"
-      icon  = "/custom/icons/intellij.svg"
-      build = "251.26927.53"
-    }
-
-    "PY" = {
-      name  = "PyCharm"
-      icon  = "/custom/icons/pycharm.svg"
-      build = "251.23774.211"
-    }
+    "GO" = { build = "261.22158.291" }
+    "PY" = { build = "261.22158.340" }
+    # Add entries for other IDEs as needed.
   }
+
+  options = ["GO", "PY"] # Must match the keys in ide_config.
 }
 ```
 
@@ -108,7 +115,7 @@ module "jetbrains" {
 module "jetbrains_pycharm" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
   agent_id = coder_agent.main.id
   folder   = "/workspace/project"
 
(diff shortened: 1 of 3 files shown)


registry/coder/modules/jetbrains/main.tf
```diff
@@ -125,95 +125,116 @@ variable "download_base_link" {
...
+    and the provided build numbers are used directly — useful for
+    air-gapped environments or pinning specific versions.
+
+    Each value must contain:
+    - name:  Display name of the IDE (e.g. "GoLand").
+    - icon:  Path or URL to the IDE icon (e.g. "/icon/goland.svg").
+    - build: Full build number (e.g. "253.28294.337").
+
     Example:
     {
```
matifali, line 148: **P1** Validation checks `ide_config ⊇ options` but not `ide_config ⊇ default`, allowing a crash when `default` includes IDE codes absent from a partial `ide_config`. *(Contract Auditor P1)*

> Scenario that passes validation but crashes at plan:
> ```tf
> default    = ["GO", "IU"]
> options    = ["GO"]
> ide_config = {
>   "GO" = { name = "GoLand", icon = "/icon/goland.svg", build = "999.99999.999" }
> }
> ```
> Validation passes: `{"GO"} ⊇ {"GO"}` ✓. But `selected_ides = toset(["GO", "IU"])` (from `var.default`), so `var.ide_config["IU"]` → key-not-found error.

This was masked before because `ide_config` defaulted to all 9 IDEs. Now that it accepts partial maps, a second validation is needed:
```hcl
validation {
  condition = var.ide_config == null || alltrue([
    for code in var.default : contains(keys(var.ide_config), code)
  ])
  error_message = "ide_config must contain entries for all IDE codes in var.default."
}
```

registry/coder/modules/jetbrains/main.tf
```diff
@@ -125,95 +125,116 @@ variable "download_base_link" {
...
-      "CL" = { name = "CLion", icon = "/icon/clion.svg", build = "253.29346.141" },
-      "GO" = { name = "GoLand", icon = "/icon/goland.svg", build = "253.28294.337" },
-      "IU" = { name = "IntelliJ IDEA", icon = "/icon/intellij.svg", build = "253.29346.138" },
+      "GO" = { name = "GoLand", icon = "/icon/goland.svg", build = "261.22158.291" },
+      "IU" = { name = "IntelliJ IDEA", icon = "/icon/intellij.svg", build = "261.22158.277" },
     }
   EOT
   type = map(object({
```
matifali, line 153: **P2** `ide_config` forces users to redundantly provide `name` and `icon` when they only care about pinning a build number. *(Product Reviewer P2, Edge Case Analyst P2)*

> The module already has `local.ide_metadata` with canonical name/icon for every IDE code. Yet `ide_config` requires all three fields. For the dominant air-gapped use case — “I just want to pin builds” — users must copy-paste 9 lines of boilerplate the module already knows.

A simpler shape like `map(object({ build = string, name = optional(string), icon = optional(string) }))` with fallback to `local.ide_metadata` would cut the config footprint significantly. Not a blocker for this PR but worth considering as a fast follow-up — it would make the README example much less intimidating.

registry/coder/modules/jetbrains/main.tf
```diff
@@ -125,95 +125,116 @@ variable "download_base_link" {
...
     name  = string
     icon  = string
     build = string
```
matifali, line 154: **P2** `ide_config` subset without explicit `options` is a likely user footgun. *(Edge Case Analyst P2)*

> `var.options` defaults to all 9 IDE codes. The natural user intent for air-gapped pinning is to provide only the IDEs they want — e.g. `ide_config = { "GO" = {...}, "PY" = {...} }`. This fails validation because `options` still defaults to all 9. The error message doesn't tell users they also need to set `options`.

This validation is pre-existing, but this PR makes it much more likely to be hit by encouraging `ide_config` usage. The README examples all set `options` alongside `ide_config` which masks the issue. Consider either:
- Updating the error message to mention `options`
- Adding a note in the `ide_config` variable description

registry/coder/modules/jetbrains/jetbrains.tftest.hcl
```diff
@@ -1,53 +1,3 @@
...
-    agent_id = "foo"
-    folder   = "/home/coder"
-  }
-
-  assert {
-    condition     = length(var.ide_config) == length(var.expected_ide_config)
-    error_message = "Test configuration mismatch: 'var.ide_config' in main.tf has ${length(var.ide_config)} items, but 'var.expected_ide_config' in the test file has ${length(var.expected_ide_config)} items. Please update the test file's global variables block."
-  }
-
-  assert {
```
matifali, line 31: **Nit** `uses_ide_config_for_multiple_ides` doesn't assert `json_data == null` for either IDE. *(Test Auditor P2)*

> The single-IDE variant `uses_ide_config_when_set` explicitly asserts `json_data == null`. The multi-IDE variant checks `name` and `build` but never `json_data`. If a regression causes `json_data` to be populated when `ide_config` is set, only the single-IDE test would catch it.

registry/coder/modules/jetbrains/README.md
```diff
@@ -14,7 +14,7 @@ This module adds JetBrains IDE buttons to launch IDEs directly from the dashboar
 module "jetbrains" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/jetbrains/coder"
-  version  = "1.3.1"
+  version  = "1.4.0"
```
matifali, line 17: **Nit** The pinned-versions example lists all 9 IDEs but the typical air-gapped user probably pins only 2–3. *(Product Reviewer Obs)*

> The wall of 9 entries makes the feature look heavier than it is.

Consider showing 2–3 IDEs with a comment `# Add entries for other IDEs as needed` and optionally a collapsible `<details>` block with the full list.

matifali (commented):
Clean fix for a real air-gapped pain point. The two-mode design (`ide_config` null vs set) is intuitive and the validation guardrails are well thought out.

1 P1, 2 P2, 2 nits across 5 inline comments. The P1 is a validation gap that can crash at plan time.

This review contains a finding that may need attention before merge.

35C4n0r (approved):
LGTM!

PR #840, opened 2026-04-13
feat(coder/modules/boundary): add agent-firewall module

35C4n0r:
## Description

Extracts boundary installation and wrapper logic into a standalone `coder/agent-firewall` module, decoupling it from `agentapi`.

### Why

Boundary is currently embedded inside `agentapi` (`scripts/boundary.sh`) and duplicated in `claude-code`. This couples network isolation to the AI/Tasks stack, but boundary is a general-purpose primitive — users running a plain agent with no agentapi or tasks should be able to use it too.

### What this adds

`registry/coder/modules/agent-firewall/` — a new first-class module that:

* Installs boundary via one of three strategies:
  1. `coder boundary` subcommand (default, zero-install)
  2. Direct binary from release (`use_agent_firewall_directly = true`)
  3. Compiled from source (`compile_agent_firewall_from_source = true`)
* Ships a comprehensive [default allowlist config](registry/coder/modules/agent-firewall/config.yaml.tftpl) (Anthropic, OpenAI, VCS, package managers, cloud platforms, etc.)
* Auto-fills the Coder deployment domain via `data.coder_workspace.me.access_url`
* Supports inline config (`agent_firewall_config`) or external file (`agent_firewall_config_path`), mutually exclusive with cross-variable validation
* Creates a wrapper script at `$HOME/.coder-modules/coder/agent-firewall/scripts/agent-firewall-wrapper.sh`
* Strips `CAP_NET_ADMIN` from the coder binary (copies to `coder-no-caps`) to allow execution inside network namespaces without `sys_admin`
* Supports `pre_install_script` / `post_install_script` hooks
* Exposes `agent_firewall_wrapper_path`, `agent_firewall_config_path`, and `scripts` outputs for script coordination
* No env vars exported — everything is output-only

### Usage

```tf
module "agent-firewall" {
  source   = "registry.coder.com/coder/agent-firewall/coder"
  version  = "0.0.1"
  agent_id = coder_agent.main.id
}
```

Works standalone with any agent — no agentapi dependency required.

### Testing

* 8 Terraform plan tests (`agent-firewall.tftest.hcl`): default outputs, compile from source, use directly, custom hooks, custom module directory, inline config, external config path, mutual exclusion validation
* TypeScript integration tests (`main.test.ts`): state verification, coder subcommand happy path, inline config, config path skip, custom hooks, env var absence, wrapper execution, idempotent installation

## Type of Change

- [X] New module

## Module Information

[...]

PR #842, opened 2026-04-13
feat(registry/coder/modules/coder-utils): make install_script and start_script optional

### LINKED ISSUE #848: Rework coder-utils module: make start_script optional, install_script required
## Summary

The `coder-utils` module currently requires `start_script` but gives `install_script` a default of `null`. In practice this is backwards — most consumers always need an install step but may not need a persistent start process. The module should be reworked so that `install_script` is the only required script and `start_script` becomes optional.

## Current Behavior

- `install_script` has `default = null` (optional).
- `start_script` is required (no default).
- `module_dir_name` sets a directory under `$HOME/<name>`.
- Four separate outputs (`pre_install_script_name`, `install_script_name`, `post_install_script_name`, `start_script_name`) are always populated, even when the corresponding script is not created.

### 1. Flip required/optional for install vs start
- `install_script` — **required** (remove default). The `coder_script` resource is always created (no `count`).
- `start_script` — **optional** (`default = null`). Conditionally created with `count = var.start_script != null ? 1 : 0`.

### 2. Rename `module_dir_name` → `module_directory`
Give callers direct control over the full working directory path instead of only a subdirectory name. This is a breaking variable rename.

### 3. Consolidate outputs into a single map
Replace the four individual outputs with a single `script_names` map:
```hcl
output "script_names" {
  value = {
    pre_install  = var.pre_install_script != null ? local.pre_install_script_name : ""
    install      = local.install_script_name
    post_install = var.post_install_script != null ? local.post_install_script_name : ""
    start        = var.start_script != null ? local.start_script_name : ""
  }
}
```
Empty string (`""`) for scripts that are not provided, rather than emitting a name for a resource that does not exist.

### 4. Move sync dependency resolution into `locals`
Introduce `install_sync_deps` and `start_sync_deps` locals to compute the correct `coder exp sync want` arguments, removing the duplicated conditional logic from inline script templates.

### 5. Add `mkdir -p` to every script block
With `install_script` now potentially the first script to run (no `pre_install_script`), each script block should ensure the module directory exists via `mkdir -p`.

## Breaking Changes

- `install_script` becomes required — callers that omitted it will need to provide a value.
- `start_script` becomes optional — callers relying on it being required should be unaffected (providing a value still works).
- `module_dir_name` renamed to `module_directory`.
- Four individual outputs replaced by a single `script_names` map output — callers referencing `module.coder_utils.install_script_name` must change to `module.coder_utils.script_names.install`, etc.

## Open Questions

- **Semver**: These are breaking changes. Should the version bump be `2.0.0` instead of `1.1.0`?
- **Missing newline at EOF** in `main.tf` should be fixed.

## Related

- PR: https://github.com/coder/registry/pull/842

35C4n0r:
## Description

Rework the `coder-utils` module to make `start_script` optional while keeping `install_script` required. Adds `mkdir -p` to every script block, updates the module directory path to a namespaced location, moves all sync dependency resolution into locals, and bumps Terraform to `>= 1.9`.

## Type of Change

- [x] Feature/enhancement

## Module Information

**Path:** `registry/coder/modules/coder-utils`
**New version:** `v1.1.0`

`install_script` is now required (previously had `default = null`). Callers that omitted it will need to provide a value. `start_script` now defaults to `null` (previously required). The module directory path changed from `$HOME/<name>` to `$HOME/.coder-modules/coder/coder-utils/<name>`.

## Testing & Validation

- [x] Tests pass (`terraform test` — 7 passed, 0 failed)
- [x] Code formatted
- [x] Changes tested locally

### `main.tf`
- `install_script`: now required (no default). Resource always created (no `count`).
- `start_script`: added `default = null`. Resource conditionally created with `count`.
- All four script blocks now include `mkdir -p` to ensure the module directory exists.
- `module_dir_path` updated to `$HOME/.coder-modules/coder/coder-utils/<module_dir_name>`.
- Sync deps moved to locals (`install_sync_deps`, `start_sync_deps`). Post-install always waits for install. Start waits for install (+ post_install when present). Install waits for pre_install when present.
- `install_script_name` output is always set. Other outputs return `""` when variable is not provided.
- Required Terraform version bumped to `>= 1.9`.

### `main.tftest.hcl`
- Updated for required `install_script` and optional `start_script`.
- Added `test_install_syncs_with_pre_install` and `test_start_syncs_with_post_install` to verify sync dependency chains.

### `main.test.ts`
- `install_script` added back to required variables.

### `README.md`
- Version bumped to `1.1.0`.
- Execution order updated: install is required, start is optional.

## Related Issues

Closes: #848

---
> Generated with [Coder Agents](https://coder.com)


matifali:
## Deep Review

The core goal — making `install_script` and `start_script` optional via `count` gating, matching the existing pattern for `pre_install_script` / `post_install_script` — is sound. The sync dependency chain logic (`post_install_sync_deps`, `start_sync_deps`) is correct and handles all combinations gracefully. Tests are thorough with good coverage of edge cases.

A few things to address before merging:

---

### 1. Undocumented breaking change: `module_dir_path` moved

This is the most significant concern. The PR silently changes `module_dir_path` from:

```hcl
module_dir_path = "$HOME/${var.module_dir_name}"
```

to:

```hcl
module_dir_path = "$HOME/.coder-modules/coder/coder-utils/${var.module_dir_name}"
```

This changes the on-disk path for **all** scripts, logs, and artifacts. It's not mentioned anywhere in the PR description. If intentional, it should be called out explicitly and justified. If not, it should be reverted from this PR and handled separately.

---

### 2. PR description says outputs return `null` — they return `""`

The description states:
> "All four outputs now return `null` when the corresponding variable is not provided."

But the actual code (and all test assertions) return `""` (empty string). Looks like this was fixed after the Copilot review but the description wasn't updated.

---

### 3. Missing newline at end of `main.tf`

The diff ends with `\ No newline at end of file`. Trivial fix.

---

### 4. Versioning: `1.1.0` but marked as breaking

The README bumps the version to `1.1.0`, but the PR description checks `[x] Yes` for breaking change. If it's truly breaking (and the `module_dir_path` change makes it so), semver convention would be `2.0.0`. If the argument is that existing callers who provide these variables are unaffected, then uncheck the breaking change box.

---

### 5. Sync dependency chain — verified correct

I traced through all combinations:

| pre_install | install | post_install | start | `post_install_sync_deps` | `start_sync_deps` |
|---|---|---|---|---|---|
| Y | Y | Y | Y | install | install + post_install |
| N | Y | N | Y | — | install |
| Y | N | Y | Y | pre_install | post_install |
| N | N | N | Y | — | null |
| Y | N | N | Y | — | pre_install |

All correct. The chain degrades gracefully when any middle step is absent.

---

### 6. `mkdir -p` added to `post_install_script` and `start_script` — good

With `install_script` now optional, the directory may not exist when later scripts run. The `mkdir -p` addition is the right fix.

---

### 7. Minor: test coverage gap

No test for `post_install_only` (just `post_install_script`, no other scripts). This would verify `post_install_sync_deps` resolves to `null` when both `pre_install_script` and `install_script` are absent.

---

| Area | Status |
|---|---|
| Core optionality logic | ✅ Good |
| Sync dependency chain | ✅ Correct |
| Test coverage | ✅ Good (minor gap) |
| `mkdir -p` fix | ✅ Good |
| `module_dir_path` change | ⚠️ Undocumented — needs discussion |
| PR description accuracy | ⚠️ Needs update (`null` → `""`) |
| Version bump vs breaking flag | ⚠️ Inconsistent |
| Missing newline at EOF | 🔧 Trivial fix |

registry/coder/modules/coder-utils/main.tf
```diff
@@ -134,7 +155,9 @@ resource "coder_script" "post_install_script" {
     set -o pipefail
 
     trap 'coder exp sync complete ${local.post_install_script_name}' EXIT
-    coder exp sync want ${local.post_install_script_name} ${local.install_script_name}
+    %{if local.post_install_sync_deps != null~}
+    coder exp sync want ${local.post_install_script_name} ${local.post_install_sync_deps}
+    %{endif~}
     coder exp sync start ${local.post_install_script_name}
```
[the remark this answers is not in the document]
matifali: This was addressed in a follow-up commit — both `post_install_script` and `start_script` now include `mkdir -p ${local.module_dir_path}` near the top.

registry/coder/modules/coder-utils/main.tf
```diff
@@ -171,20 +193,20 @@ resource "coder_script" "start_script" {
...
   description = "The name of the install script for sync."
-  value       = local.install_script_name
+  value       = var.install_script != null ? local.install_script_name : null
 }
 
 output "post_install_script_name" {
   description = "The name of the post-install script for sync."
-  value       = local.post_install_script_name
+  value       = var.post_install_script != null ? local.post_install_script_name : null
 }
```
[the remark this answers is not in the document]
matifali: This was addressed — outputs now return `""` (empty string) instead of `null`, which is safer for downstream callers that might concatenate or pass these values. The PR description still says `null` though and should be updated to match the actual behavior.

registry/coder/modules/coder-utils/main.tf
```diff
@@ -58,14 +59,14 @@ locals {
...
-  encoded_start_script        = base64encode(var.start_script)
+  encoded_start_script        = var.start_script != null ? base64encode(var.start_script) : ""
 
   pre_install_script_name  = "${var.agent_name}-pre_install_script"
   install_script_name      = "${var.agent_name}-install_script"
   post_install_script_name = "${var.agent_name}-post_install_script"
   start_script_name        = "${var.agent_name}-start_script"
 
-  module_dir_path = "$HOME/${var.module_dir_name}"
+  module_dir_path = "$HOME/.coder-modules/coder/coder-utils/${var.module_dir_name}"
```
DevelopmentCats: Ultimately this is a NIT and not a blocker.

I know we were debating on where we actually wanted all of this to live but did we finally decide that this is what we wanted over anything else?

I only ask because `coder-registry/` comes to mind since we could isolate all of module related files into that one folder since they derive from here.

matifali: there is also `CODER_SCRIPT_BIN_DIR`.

35C4n0r: > I only ask because coder-registry/ comes to mind since we could isolate all of module related files into that one folder since they derive from here.

like $HOME/coder-registry/author/module_name/* ?

also we also want this to be configurable, so I'll add a variable here as I'm planning to do with boundary, tasks , and other ai module.
```
locals{
    resolved_base_module_dir = var.base_module_dir ? var.base_module_dir : "$HOME/coder-registry"
    module_dir_path = "${local.resolved_base_module_dir}/author/${var.module_dir_name}"
}
```

@DevelopmentCats @matifali thoughts ?

35C4n0r: oof, that make author hardcoded, and other users won't be able to use it.

35C4n0r: >I know we were debating on where we actually wanted all of this to live but did we finally decide that this is what we wanted over anything else?

[thread](https://www.notion.so/Refactor-Make-Agent-Modules-Independent-of-Tasks-AgentAPI-30bd579be5928058b7bff73ebac753dc?d=31fd579be59280e09ee1001c80ba18bf&source=copy_link#314d579be59280a69bb0c9631ff27938)

35C4n0r: Ok, I've made it so that path will be provided by the parent module, given that this is a utils module - we make no defaults

DevelopmentCats (approved):
LGTM provided we have tested this, but if not I can finagle something together to test it all

35C4n0r:
> LGTM provided we have tested this, but if not I can finagle something together to test it all

Currently testing it in the new boundary module. Will merge this right before #840.
Thanks for the review @DevelopmentCats

matifali (approved):

matifali:
Pushed 6678b04 on top of this branch to add two optional variables so downstream modules (starting with `claude-code`) can brand the scripts this module emits:

- `display_name_prefix` (string, default `""`) — prepends `"<prefix>: "` to every `coder_script.display_name`.
- `icon` (string, default `null`) — sets `icon` on every `coder_script`.

Existing tests unchanged (defaults preserve the old behavior). Two new cases in `main.tftest.hcl` cover both vars. README updated with a "Customizing Script Display" section.

Also fixed a stale `module_dir_name` reference in the README example (the variable was renamed to `module_directory` earlier on this branch).

Let me know if you'd rather I split this into a follow-up PR once this one merges.

registry/coder/modules/coder-utils/main.tf
```diff
@@ -49,46 +49,67 @@ variable "agent_name" {
...
+variable "module_directory" {
+  type        = string
+  description = "The module's working directory for scripts and logs."
+}
+
+variable "display_name_prefix" {
+  type        = string
+  description = "Prefix for each coder_script display_name. Example: setting 'Claude Code' yields 'Claude Code: Install Script', 'Claude Code: Pre-Install Script', etc. When unset, scripts show as plain 'Install Script'."
+  default     = ""
+}
```
35C4n0r, line 61: @matifali should we just reuse upper(var.agent_name) instead ?

matifali: that's a good idea. is agnet name used anywhere else?

35C4n0r: the agent_name is used to make unique script names for `coder exp sync` purposes.

matifali: I will leave both as the name I need for display is a free-form Title case and not a slug.

matifali:
One more addition on this branch (`178c609`): new `output "scripts"` that returns the run-ordered, filtered list of `coder exp sync` names this module actually creates, with absent scripts omitted.

This is generic to anything wrapping `coder-utils`, so it belongs here rather than being re-derived in every consumer. It unblocks a simplification in #859 (claude-code) where the same filter was being done locally. 3 new tftest cases cover all-scripts-present, install-only, and install+post.

PR #846, opened 2026-04-16
fix(claude-code): don't pass --dangerously-skip-permissions in auto mode

bdevore17:
## Summary
- The task reporting code path in `start.sh` unconditionally adds `--dangerously-skip-permissions` regardless of the configured `permission_mode`
- Workspaces using `permission_mode = "auto"` get both `--permission-mode auto` and `--dangerously-skip-permissions`, which defeats the purpose of auto mode
- Only add `--dangerously-skip-permissions` when no explicit permission mode is set (backward compat) or when `bypassPermissions` is chosen

## Test plan
- [x] Added `claude-continue-resume-task-session-auto-mode` test verifying `--dangerously-skip-permissions` is NOT passed when `permission_mode=auto`
- [x] Updated `claude-auto-permission-mode` test to assert `--dangerously-skip-permissions` is absent
- [x] Existing `claude-continue-resume-task-session` test still verifies default (no permission_mode) includes `--dangerously-skip-permissions`

### DIFF
--- a/registry/coder/modules/claude-code/main.test.ts
+++ b/registry/coder/modules/claude-code/main.test.ts
@@ -180,6 +180,8 @@ describe("claude-code", async () => {
       "cat /home/coder/.claude-module/agentapi-start.log",
     ]);
     expect(startLog.stdout).toContain(`--permission-mode ${mode}`);
+    // Auto mode should not include --dangerously-skip-permissions
+    expect(startLog.stdout).not.toContain("--dangerously-skip-permissions");
   });
 
   test("claude-model", async () => {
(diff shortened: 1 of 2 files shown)


bdevore17:
@35C4n0r please review

DevelopmentCats:
@bdevore17 

Have you tested this with tasks yet? I only ask because the reason we unconditionally added `--dangerously-skip-permissions` when task mode is enabled, was to allow for the agent to work autonomously in the background, as previously claude-code would stop and require user input which slightly defeated the purpose of running this in task mode.

bdevore17:
@DevelopmentCats yes i have. This still works as intended. The only difference is it first respects the permission_mode in the template. Right now, if you set permission_mode to `auto`, it doesn't work at all because the bypass permissions flag overrides it.

DevelopmentCats:
> @DevelopmentCats yes i have. This still works as intended. The only difference is it first respects the permission_mode in the template. Right now, if you set permission_mode to `auto`, it doesn't work at all because the bypass permissions flag overrides it.

Perfect! Thank you so much for the testing and the contribution. I will go ahead and put this through some rounds of testing just to confirm since this module is active, and then we can get this merged! 😸

DevelopmentCats:
Closing because of Tasks/Agents Refactor

Issue #82, opened 2023-10-18
feat: allow execution ordering of coder modules


djarbz:
How can we specify the order in which modules run?

For example, I want the Personalize module to run after my Dotfiles have been fully cloned as I have a personalization script in my dotfiles.

matifali:
@DevelopmentCats do you have any ideas here? I think this depends on https://github.com/coder/coder/issues/10352

35C4n0r:
Users can now use: `coder exp sync` to coordinate their scripts.
Docs: https://coder.com/docs/admin/templates/startup-coordination/usage
PR: https://github.com/coder/coder/pull/20758

PR #849, opened 2026-04-17
feat: add tailscale module

dy-ma:
## Description

Adds a Tailscale module that installs Tailscale and joins the workspace to a tailnet on startup.

Supports:
- OAuth client credentials flow (default) or a pre-generated auth key
- Kernel and userspace networking (auto-detected, or set explicitly)
- Tailscale SSH, subnet route advertising, MagicDNS, ephemeral/persistent nodes
- SOCKS5 and HTTP proxies in userspace mode

Tailscale seemed like an obvious fit for a Coder module and I was surprised one didn't already exist. I needed it for my own deployments so I built it.

This is my first open-source contribution. I've done my best to follow the contribution guidelines, but please tell me if I've done something horribly wrong.

## Type of Change

- [x] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/dy-ma/modules/tailscale`  
**New version:** `v1.0.0`  

## Template Information

**Path:** `registry/[namespace]/templates/[template-name]`

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

None

Generated with Claude Code using Claude Sonnet 4.5


35C4n0r:
@dy-ma can you fix the failing test ?

dy-ma:
Fixed the ShellCheck failure - `sudo` doesn't affect redirects, so I changed line 96 from `&> /tmp/tailscaled.log` to use `sudo tee` instead.

The fix passes ShellCheck locally. Ready for another review when you have a chance.

matifali:
Hi @dy-ma, can you update the module to conform with instructions in `AGENTS.md`?

dy-ma:
As far as I can see, the only non-conformance was the Tailscale icon being referanced via CDN link instead of local path. I initially wasn't sure of the correct approach for adding new icons, but since I had already added the icon to `.icons/` for the README, I've updated the module to use the local path as well. 

Also, I wasn't aware of the AI attribution policy, so I edited my comment. I appreciate the patience, as this is my first open source contribution.

registry/dy-ma/modules/tailscale/main.tf
```diff
@@ -0,0 +1,219 @@
...
+      source  = "coder/coder"
+      version = ">= 2.5"
+    }
+  }
+}
+
+data "coder_workspace" "me" {}
+
+locals {
+  icon_url  = "${path.module}/../../../../.icons/tailscale.svg"
```
matifali: This will not work. Coder can only show [bundled icons](https://coder.com/docs/admin/templates/extending-templates/icons#bundled-icons) or something accessible over the network.

registry/dy-ma/modules/tailscale/main.tf
```diff
@@ -0,0 +1,219 @@
...
+    ACCEPT_DNS          = var.accept_dns
+    ACCEPT_ROUTES       = var.accept_routes
+    ADVERTISE_ROUTES    = join(",", var.advertise_routes)
+    SSH                 = var.ssh
+    EXTRA_FLAGS         = var.extra_flags
+    STATE_DIR           = var.state_dir
+  })
+  run_on_start = true
+  run_on_stop  = false
+}
```
matifali: You could try using the [`coder-utils`](https://registry.coder.com/modules/coder/coder-utils) module to manage installation and start the service. 

This would also help comply with the suggested logs storage preference.

registry/dy-ma/modules/tailscale/main.tf
dy-ma: I see. The guidelines mention the README icon must be a local path. There isn't a bundled Tailscale icon. Is the correct approach for me to add the icon in .icons, but only reference it for the README, and use a network call for the module?

matifali:
`.icons` is for the registry.coder.com and this repo.<br>But for the scripts to show this icon on real coder deployments we need the icon refrenced from [built-in icons](<https://github.com/coder/coder/tree/main/site/static/icon>) at `coder/coder` or a network path.

dy-ma:
I split `run.sh` into `install.sh` and `start.sh`, and used the coder_utils module to orchestrate them, and added a module directory for logging (I believe this is what you were mentioning by "log storage preference"?)

I considered using a post-install configuration, but in order to configure tailscale, the daemon needs to be running first, and I figure I'd want the flags to be reapplied on restart anyways. So I kept it to just a `start.sh`.

dy-ma:
Upon further testing, I ran into an apt lock race condition between this module and KasmVNC, as both run apt-get update concurrently at boot, and whichever runs second fails with E: Could not get lock /var/lib/apt/lists/lock. 

I looked into solutions and found [coder/coder#22365](https://github.com/coder/coder/issues/22365) and [coder/registry#82](https://github.com/coder/registry/issues/82). The experimental `coder exp sync` feature could coordinate ordering but requires the template author to wire it up. It doesn't seem like I can have the module self-coordinate using it. 

I'm currently testing a few approaches for handling this in `install.sh`: 
- Wrapping `apt-get` in `/usr/local/bin` with `APT::Lock::Timeout` so the timeout applies even through `sudo` (similar to [tailscale/tailscale#12291](https://github.com/tailscale/tailscale/issues/12291)
- Doing the apt repo setup directly instead of piping to the official install script, so we can control the `apt-get` calls ourselves. But this requires branching the install methods for each distro, and introducing opportunities for drift between this module and the official installation methods.
- Polling for the apt lock to be released between retries. 

Happy to go with whatever approach you think fits best here, or if there's a preferred pattern for modules, please let me know.

DevelopmentCats:
Nice work for a first contribution, this module covers a lot of ground.

## apt lock race condition

On the apt lock issue you ran into with kasmvnc - the apt-get wrapper in `install.sh` should come out. The way we handle script ordering between modules in Coder is through `coder exp sync`, which `coder-utils` already sets up for you. Multiple `coder_script` resources run in parallel by default, so the real fix is letting template authors order modules explicitly. You need to:

1. Expose a `pre_install_script` variable and pass it through to `coder-utils` so template authors can order other modules before this one
2. Expose the `scripts` output so other modules/scripts can wait on this one to finish

That way if a template uses both tailscale and kasmvnc, the author can wire them so they don't run apt concurrently. Check how `claude-code` does this in its `main.tf`.

With that, `install.sh` can just be:

```bash
#!/usr/bin/env bash
set -euo pipefail

log() { echo "[tailscale-install] $*" >&2; }
has() { command -v "$1" &> /dev/null; }

if has tailscale; then
  log "Tailscale already installed ($(tailscale version 2>/dev/null | awk 'NR==1{print $1}')), skipping."
  exit 0
fi

log "Installing Tailscale..."
curl -fsSL https://tailscale.com/install.sh | sh
log "Installed: $(tailscale version | head -1)"
```

Add the variable:

```tf
variable "pre_install_script" {
  description = "Custom script to run before installing Tailscale. Use this to order this module after another module's install pipeline."
  type        = string
  default     = null
}
```

Pass it through in the `coder_utils` module block:

```tf
module "coder_utils" {
  source  = "registry.coder.com/coder/coder-utils/coder"
  version = "0.0.1"

  agent_id            = var.agent_id
  module_directory    = "$HOME/.coder-modules/dy-ma/tailscale"
  display_name_prefix = "Tailscale"
  icon                = local.icon_url
  pre_install_script  = var.pre_install_script
  install_script      = file("${path.module}/install.sh")
  start_script        = local.start_script
}
```

And add the `scripts` output:

```tf
output "scripts" {
}
```

## start.sh should be a .tftpl

Since `start.sh` goes through `templatefile()`, rename it to `scripts/start.sh.tftpl`. This means:
- Bash `$` needs to be escaped as `$$` (e.g. `$${HOME}`)
- Sensitive vars (`AUTH_KEY`, `OAUTH_CLIENT_ID`, `OAUTH_CLIENT_SECRET`) should be `base64encode()`'d on the terraform side and decoded in the script with `$(echo -n '${VAR}' | base64 -d)` so special characters don't break the template
- Booleans should use `tostring()` on the terraform side

Render it in a `locals` block:

```tf
locals {
  start_script = templatefile("${path.module}/scripts/start.sh.tftpl", {
    AUTH_KEY            = base64encode(var.auth_key)
    OAUTH_CLIENT_ID     = base64encode(var.oauth_client_id)
    OAUTH_CLIENT_SECRET = base64encode(var.oauth_client_secret)
    EPHEMERAL           = tostring(var.ephemeral)
    # ... etc
  })
}
```

See `claude-code`'s `scripts/install.sh.tftpl` for the full pattern.

## Variable ordering

`auth_key`, `oauth_client_id`, `oauth_client_secret` have `sensitive` before `default` and `description` last. AGENTS.md wants: `description` → `type` → `default` → `validation` → `sensitive`.

---

> _This review was generated with the assistance of Coder Agents._

DevelopmentCats:
@dy-ma Also for the CI failing you can just run 'bun run fmt' and it will format everything and the CI will pass

dy-ma:
Thanks for the detailed review, the examples really helped.

I've applied the module ordering and templating changes.

One issue came up during `terraform test`:

> Error: Output refers to sensitive values
>
>  on main.tf line 232
> 232: output "scripts" {

`auth_key`, `oauth_client_id`, and `oauth_client_secret` are marked `sensitive = true` and passed through `templatefile()`, which taints `local.start_script` and propagates to the `scripts` output via `coder_utils`.

I see two ways to resolve this:

1. `nonsensitive()` - strip the taint at the `templatefile()` call. Simple, but feels risky.
2. `coder_env` pattern - matches how `claude-code` and `codex` handle API keys, but those keys need to persist for interactive use. OAuth credentials here are transient (exchange -> access token -> single-use auth key -> done), so leaving them in the workspace environment all session feels unnecessary.

Perhaps I'm overestimating the risk since the credentials end up on the workspace either way, but I'd appreciate some guidance on your preferred approach.

matifali:
I would suggest going with 1.

dy-ma:
Sounds good. Changes have been pushed

dy-ma:
Hi, just checking in. I believe I've addressed all the feedback from the last review. I just wanted to make sure this wasn't forgotten. Is the review still under way?

35C4n0r (approved):

DevelopmentCats:
> Hi, just checking in. I believe I've addressed all the feedback from the last review. I just wanted to make sure this wasn't forgotten. Is the review still under way?

I apologize for the delay a few of us just got back from a convention. I will go ahead and give this a final test, and merge it provided it has been approved already @dy-ma

DevelopmentCats:
Tested with `ephemeral = true` and a quick stop/start. The restarted node got MagicDNS name `apricot-gull-86-1` because the old ephemeral node was still in the tailnet as `offline, last seen 1m ago`.

Cleanup should be default-on, not opt-in. Matching on offline + same tag + same hostname is safe (describes exactly the previous instance of this workspace), and the OAuth client's tag scope already bounds the blast radius. If the client only has `auth_keys`, the list call 403s, log a line and continue. README should call out the behavior.

DevelopmentCats:
In userspace mode, `configure_proxy_env` unconditionally writes `/etc/profile.d/tailscale-proxy.sh`, setting `ALL_PROXY`/`http_proxy`/`https_proxy` for every login shell. That routes *all* outbound HTTP from common tooling (`curl`, `git`, `pip`, `npm`, `apt`) through tailscaled, not just tailnet-bound traffic. Suggest gating behind a `var.export_proxy_env` boolean defaulting to `true` — preserves current behavior, lets operators with mixed workloads opt out.

DevelopmentCats:
`nonsensitive(base64encode(var.oauth_client_secret))` at `main.tf:21` plus coder-utils rendering `start_script` to disk leaves the base64-encoded secret in `~/.coder-modules/dy-ma/tailscale/scripts/start.sh` (mode `0755`), recoverable via `base64 -d`. base64 is transport encoding, not protection. The chmod fix belongs in coder-utils, but the threat model is worth a README note.

DevelopmentCats:
`output.hostname` returns `local.hostname` (what was requested), not the MagicDNS name Tailscale actually assigned — which can diverge when the stale-node collision above happens. Worth one line in the README clarifying it's "the hostname this module asked Tailscale to register."

DevelopmentCats:
@dy-ma I tested everything and it works, but I left a few comments above worth addressing before merging

dy-ma:
@DevelopmentCats Wow, I really appreciate the review. I certainly could have tested this module better. I'll get on those fixes asap. 
Just checking about your third note there: are you saying the long-term fix is for coder-utils to restrict the permissions of the script, something like `chmod 600` or `chmod 700`. But this fix belongs in coder-utils, so I'll just leave a note in the README?

dy-ma:
Sorry or the noisy commit history. I ran my test workflow against a live workspace, and I thought I had to push each time to test. I've since learned I can use a local path source in my template to iterate without publishing, and will try that or future work. But, as far as I can tell, I've addressed the comments:

- [X] **Stale node cleanup** - default-on; uses the OAuth access token to list devices and delete any offline nodes matching the current hostname before `tailscale up`. 403 (insufficient scope) is caught and skipped with a log line. One subtle issue: the Tailscale API returns `"online": null` for offline nodes, not `false`. README documents the behavior and required scopes (`devices:core` write for your tag)
- [X] **`export_proxy_env`opt-out** - added as a bool (default `true`); `configure_proxy_env` gates on it so operators can opt out of the proxy env vars being written to `/etc/profile.d/`.
- [X] **README** updated to note security implications of OAuth, and the hostname divergence.

DevelopmentCats:
@dy-ma, just wanted to follow up on some behavior I'm seeing during testing.

Did you run through the full stop/start cycle? Basically starting a workspace, stopping it, starting it again, and then doing a restart?

Every time I test it, the Tailscale entry gets duplicated with an incrementing number at the end each time the workspace is stopped and started back up. This is the UX issue I was pointing out, and I think it's worth resolving before we merge.

dy-ma:
@DevelopmentCats, That's strange. I did get that issue but I thought I resolved it. I'll run it again a few times and see.

dy-ma:
Hi @DevelopmentCats, I got the chance to test again, and I'm not able to reproduce the issue.

I tested both persistent and ephemeral environments (server on ec2, pod on kubernetes). Both were able to restart and resume the same tailnet identity. Can I ask if you are testing with the latest code, and on what platform?

matifali:
@DevelopmentCats @35C4n0r is this mergeable?

matifali (approved):

DevelopmentCats (approved):
I have tested this on my end again and everything works as intended. Thank you for your contribution @dy-ma

PR #845, opened 2026-04-15
feat: add oneclaw module for 1Claw MCP integration

kmjones1979:
## Summary

- Adds the **kmjones1979** namespace and **oneclaw** module
- Ported from [1clawAI/1claw-coder-workspace-module](https://github.com/1clawAI/1claw-coder-workspace-module)
- Provides vault-backed secrets and MCP server configuration for AI coding agents (Cursor, Claude Code, etc.) in Coder workspaces
- Three provisioning modes: Terraform-native, shell bootstrap, and manual

## Files added

All under `registry/kmjones1979/`:

- **Namespace**: `README.md`, `.images/avatar.png` (from GitHub)
- **Module** (`modules/oneclaw/`): `main.tf`, `variables.tf`, `outputs.tf`, `README.md`
- **Scripts**: `scripts/provision.sh`, `scripts/bootstrap.sh`, `scripts/setup.sh`
- **Tests**: `main.tftest.hcl` (5 plan runs), `main.test.ts` (5 bun tests)

No shared repo files were modified.

## Test plan

- [x] `terraform validate` passes
- [x] `terraform test` passes (Terraform 1.14 via Docker)
- [x] `bun test main.test.ts` passes (5/5)
- [x] ShellCheck passes on all `.sh` files
- [x] `bun run fmt` applied

---

Generated with Cursor using Claude

Made with [Cursor](https://cursor.com)

PR #857, opened 2026-04-22
feat: add oneclaw module for 1Claw MCP integration

kmjones1979:
## Summary

Resubmission of #845 (closed for structural reasons) with the reviewer feedback addressed and additional security hardening of the bootstrap flow.

Adds the `kmjones1979` namespace and the `oneclaw` module, which provides vault-backed secrets and MCP server wiring for AI coding agents (Cursor, Claude Code) in Coder workspaces.

### Structural (addresses @DevelopmentCats review)

The reviewer asked that the module follow the standard schema used in the `coder/` namespace:

> Generally in almost all cases you would just have: `main.tf`, `README.md`, `main.test.ts`, `main.tftest.hcl` and script files which amount to maybe one or two scripts.

Done:

- `variables.tf` and `outputs.tf` are consolidated into `main.tf`.
- `scripts/bootstrap.sh` and `scripts/setup.sh` are merged into a single `scripts/run.sh` executed by one `coder_script`.
- The Terraform-native provisioning mode (`scripts/provision.sh`, `null_resource.provision`, `master_api_key`) is removed. That mode relied on a `local-exec` provisioner writing a state file to the provisioner's cwd, which is ephemeral inside Coder template provisioners and cannot round-trip credentials into `coder_env`. Two modes remain: **bootstrap** (recommended) and **manual**.

Final tree:

```
registry/kmjones1979/
├── README.md
├── .images/avatar.png
└── modules/oneclaw/
    ├── README.md
    ├── main.tf
    ├── main.test.ts
    ├── main.tftest.hcl
    └── scripts/run.sh
```

### Security hardening for the `1ck_` human bootstrap key

The `1ck_` human API key is privileged (can create and destroy vaults in the caller's 1Claw account), so the module goes out of its way to make sure it cannot be recovered from the workspace after bootstrap:

1. The key is delivered to the script as a **sensitive `coder_env` variable** (`_ONECLAW_HUMAN_API_KEY`) rather than via `templatefile()` substitution. As a result, the literal key never appears in the rendered script body that lives in Terraform state or in the Coder agent's `/tmp/coder-agent.log`. The rendered script only shows `HUMAN_KEY="\${_ONECLAW_HUMAN_API_KEY:-}"`.
2. The key is sent to the 1Claw auth endpoint via `curl --data-binary @-` from stdin, so it never appears in `ps aux` / `/proc/<pid>/cmdline`.
3. `HUMAN_KEY` and `_ONECLAW_HUMAN_API_KEY` are unset immediately after authentication, so downstream subprocesses spawned by the script do not inherit the key.
4. Only the scoped `ocv_` agent key and the vault id are persisted to `~/.1claw/bootstrap.json` and the MCP config files.
5. `README.md` documents a post-bootstrap cleanup flow: once the state file exists, the user is instructed to set `human_api_key = ""` in their Terraform so subsequent restarts do not reference the human key at all.

## Test plan

Verified against a local Coder server (v2.31.9) running the module with real 1Claw credentials.

- [x] `terraform test` passes (4 runs, Terraform 1.14 via Docker)
- [x] `bun test main.test.ts` passes (5 tests, including an explicit assertion that the human key value is **not** embedded in the rendered script)
- [x] `shellcheck` is clean on `scripts/run.sh`
- [x] `bun run fmt` leaves the tree unchanged
- [x] Live workspace on local Coder: first boot creates vault + agent + policy and writes `bootstrap.json` + Cursor/Claude MCP configs
- [x] Live workspace restart is idempotent — the script detects the state file and skips provisioning
- [x] Post-bootstrap cleanup flow (`human_api_key = ""`, `coder update`, restart) continues to work using cached credentials
- [x] Filesystem audit after each scenario confirms the `1ck_` key value does **not** appear in any file on the workspace (state file, MCP configs, agent log, script log, shell init files, `/proc/<pid>/environ` of any Coder process)

Made with [Cursor](https://cursor.com)


kmjones1979:
Thanks for the careful review — addressed all seven items in 4d5b414.

| Copilot comment | File | Fix |
|---|---|---|
| `eval echo` on STATE_DIR enables command substitution | `scripts/run.sh:17` | Replaced with `expand_path()` helper that does pure bash string substitution (`${p//\$HOME/$HOME}` and a `case` for `~`/`~/...`). Unit-tested locally with `'/tmp/$(touch …).cfg'` and backtick payloads to confirm nothing executes. |
| `eval echo` on user-provided config paths | `scripts/run.sh:164` | Same `expand_path()` helper. |
| Vault name interpolated into inline Python | `scripts/run.sh:108-114` | Vault name is now passed via argv. While I was there I also fixed an unrelated bug shellcheck caught (SC2259): the rewrite originally used `python3 - "$NAME" <<'PYEOF'` which would have overridden the piped JSON with the heredoc; the final form uses `python3 -c '...' "$NAME" <<<"$list_response"` so argv carries the name, stdin carries the JSON, and the Python source is never interpolated. |
| Stale `master_api_key` reference in warning | `scripts/run.sh:214-216` | Updated to "Provide api_token + vault_id (manual mode), or set human_api_key (bootstrap mode)" since the master-key path was removed earlier in this PR. |
| `order` variable declared but unused | `main.tf:118-122` | Removed. `coder_script` doesn't expose an order field, so it was dead surface. |
| `data "coder_workspace_owner" "me"` unused | `main.tf:126` | Removed. |
| `_ONECLAW_HUMAN_API_KEY` lingered in `/proc/<pid>/environ` for the whole run | `scripts/run.sh:83-86` | Moved `unset _ONECLAW_HUMAN_API_KEY` to immediately after `HUMAN_KEY="${_ONECLAW_HUMAN_API_KEY:-}"` at the top of the script (line 39 now). The shell variable `HUMAN_KEY` is still scrubbed after auth completes, and the belt-and-suspenders unset of `HUMAN_KEY` after the `bootstrap` call is kept for the state-file short-circuit path. |

All checks still green: `terraform test` (4 runs), `bun test` (5 tests), `shellcheck`, `bun run fmt`.

kmjones1979:
I think all issues were fixed, tested this full flow locally and all works well. Let me know if you find anything else. Thanks for all the support on this PR.

kmjones1979:
Addressed all **8 suppressed Copilot findings** from the Aug 12 review in `0b7c957`.

| Copilot finding | Fix |
|---|---|
| JSON bodies built via string interpolation (`run.sh` vault/agent/policy/auth) | All API request bodies now serialized with Python `json.dumps()` helpers (`json_vault_create`, `json_agent_create`, `json_policy_create`, `json_auth_body`) |
| `api_call()` unreliable under `set -e` + `pipefail` | Capture curl exit code with `\|\| rc=$?` so fallbacks (e.g. vault create → list by name) work correctly |
| Missing `umask 077` before writing secrets | Added at top of install script |
| Runtime data under `~/.1claw` | Moved to `$HOME/.coder-modules/kmjones1979/oneclaw/` (bootstrap.json, logs via coder-utils); legacy `~/.1claw/bootstrap.json` auto-migrated on next start |
| Raw `coder_script` instead of `coder-utils` | Migrated to `module "coder_utils"` with `scripts/install.sh.tftpl`; exposes `output "scripts"` for downstream sync |
| Dead `agent_id_1claw` config | Manual mode still sets `ONECLAW_AGENT_ID` via `coder_env`; bootstrap mode loads agent ID from `bootstrap.json` and exports `ONECLAW_AGENT_ID` at runtime |
| (prior June review items) | Still fixed: `expand_path()`, argv-safe Python, early `_ONECLAW_HUMAN_API_KEY` unset, no dead TF vars |

**Tests run locally:**
- `terraform test` — 4/4 pass (TF 1.14)
- `bun test main.test.ts` — 5/5 pass
- `go run ./cmd/readmevalidation` — pass
- `bun run fmt` — clean

**Note:** Bootstrap no longer sets `start_blocks_login` (coder-utils does not expose it today; same pattern as `claude-code`). Bootstrap still runs on first workspace start via the install script.

DevelopmentCats (approved):
LGTM I tested everything in my own workspaces as well and it works like a charm!

Thank you very much for the contribution!!

PR #861, opened 2026-04-22
chore(claude-code)!: strip boundary, agentapi, tasks, tools

matifali:
## Summary

Reduces `claude-code` to a minimal install-and-configure module for the Claude CLI. Boundary, agentapi, task reporting, session tracking, tool allow/deny, and start orchestration are removed. Script execution is delegated to `coder-utils` v1.2.0. Starting Claude is now the caller's responsibility (template command, IDE launcher, or a custom `coder_script`).

> [!WARNING]
> This is a breaking change. See the migration guide below.

## Changes

- Delete `scripts/start.sh`. The module no longer launches Claude.
- Replace the `agentapi` sub-module with `coder-utils` v1.2.0 for install/pre/post script orchestration.
- Remove Boundary install, agentapi install, task reporting, session tracking, and tool allow/deny logic.
- Remove the npm installation path. The module now always uses the official `claude.ai/install.sh` installer.
- Render `scripts/install.sh.tftpl` once at plan time via `templatefile()` and pass the result directly to `coder-utils`, replacing the runtime base64 encode/decode wrapper and the redundant `scripts/install.sh`.
- Make `workdir` optional. When set, the module pre-creates the directory and pre-accepts the Claude Code trust dialog for it. When unset, the module installs the CLI and configures authentication only; users accept trust dialogs interactively per project.
- Scope MCP servers to Claude Code's user scope via `claude mcp add-json --scope user` so they are available across every project the workspace owner opens, instead of being tied to a single project directory.
- Drop `primaryApiKey` from the standalone-mode config writer — credentials are already exported via `coder_env`, writing the key into `~/.claude.json` had no effect.
- Rename variable `claude_api_key` → `anthropic_api_key` (env `ANTHROPIC_API_KEY`).
- Rename variable `enable_aibridge` → `enable_ai_gateway`.
- AI Gateway now sets `ANTHROPIC_AUTH_TOKEN` (not `ANTHROPIC_API_KEY`) to the workspace owner's Coder session token, matching the [AI Gateway centralized API key flow](https://coder.com/docs/ai-coder/ai-gateway/clients/claude-code#centralized-api-key).
- Fix `configure_standalone_mode` to also recognize `CLAUDE_CODE_OAUTH_TOKEN` so OAuth-only users get the onboarding bypass.
- Add `scripts` output (pass-through from `coder-utils`) so upstream modules can serialize their `coder_script` resources behind this module's install pipeline via `coder exp sync want`.
- Remove the `task_app_id` output.
[...]

PR #862, opened 2026-04-22
feat(claude-code): add telemetry input for OTEL export with workspace attribution

morganl-ant:
## Problem

Claude Code ships an OpenTelemetry exporter for token usage, tool calls, session lifecycle and errors (https://docs.anthropic.com/en/docs/claude-code/monitoring-usage), but the module exposes no first-class wiring for it. Template authors who want telemetry have to know the env var names (`CLAUDE_CODE_ENABLE_TELEMETRY`, the `OTEL_EXPORTER_OTLP_*` family) and write their own `coder_env` blocks. More importantly there is no convention for how to correlate Claude Code telemetry with Coder's own audit logs and `exectrace` records, so even when both are exported they end up as two unjoined datasets.

## Change

Adds a `telemetry` input that turns on `CLAUDE_CODE_ENABLE_TELEMETRY` and the standard OTLP exporter env vars in one place:

```tf
telemetry = {
  enabled       = true
  otlp_endpoint = "http://otel-collector.observability:4317"
  otlp_protocol = "grpc"
  otlp_headers  = { authorization = "Bearer ..." }
  resource_attributes = { "service.name" = "claude-code" }
}
```

When enabled, the module automatically appends `coder.workspace_id`, `coder.workspace_name`, `coder.workspace_owner` and `coder.template_name` to `OTEL_RESOURCE_ATTRIBUTES`. This gives a stable join key between Claude Code spans/metrics and Coder's audit log and exectrace events on `workspace_id`, so a platform team can answer "show me every shell command Claude executed in workspace X alongside the token spend for that session" without custom plumbing.

This is purely additive (`coder_env` resources behind `count`), defaults to disabled, and is independent of how Claude is launched, so it composes cleanly with the install-only direction in #861.

## Validation

- `terraform fmt`, `terraform validate`, `terraform test` (19/19) pass
- `bun test -t telemetry` (2/2) pass: env vars are set with the expected values when enabled, and absent when the input is omitted

Disclosure: I work at Anthropic on the Claude Code team.

### DIFF
--- a/registry/coder/modules/claude-code/main.tf
+++ b/registry/coder/modules/claude-code/main.tf
@@ -273,6 +273,18 @@ variable "enable_state_persistence" {
   default     = true
 }
 
+variable "telemetry" {
+  type = object({
+    enabled             = optional(bool, false)
+    otlp_endpoint       = optional(string)
+    otlp_protocol       = optional(string, "http/protobuf")
+    otlp_headers        = optional(map(string), {})
+    resource_attributes = optional(map(string), {})
+  })
+  description = "Configure Claude Code OpenTelemetry export. When enabled, sets CLAUDE_CODE_ENABLE_TELEMETRY and the standard OTEL_EXPORTER_OTLP_* environment variables. Coder workspace identifiers (coder.workspace_id, coder.workspace_name, coder.workspace_owner, coder.template_name) are automatically appended to OTEL_RESOURCE_ATTRIBUTES so Claude Code telemetry can be joined with Coder audit and exectrace logs."
+  default     = {}
+}
+
 resource "coder_env" "claude_code_md_path" {
   count    = var.claude_md_path == "" ? 0 : 1
   agent_id = var.agent_id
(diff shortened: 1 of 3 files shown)


DevelopmentCats:
@morganl-ant 

Thank you for the contribution! Everything looks good to me, but the one thing we do need in this, is for the module version to be bumped. You can use the version bump script to do so. This would be a minor version bump.

matifali:
I wonder if we provide a generic `env` block and the user can pass in arbitrary key value pairs instead of adding explicitly top-level `telemetry` block.

DevelopmentCats:
> I wonder if we provide a generic `env` block and the user can pass in arbitrary key value pairs instead of adding explicitly top-level `telemetry` block.

Yeah I thought the same for a more lightweight addition, I just thought the usage of the variable was pretty unobtrusive.

morganl-ant:
On the generic `env` block vs typed `telemetry` input: a generic env map would cover the passthrough half (`CLAUDE_CODE_ENABLE_TELEMETRY`, `OTEL_EXPORTER_OTLP_*`), but the part I think actually earns its place as a typed input is the auto-injected `OTEL_RESOURCE_ATTRIBUTES`. The module knows `data.coder_workspace.me.id` / `name` / `template_name` and `data.coder_workspace_owner.me.name` at plan time and stamps them onto every span and metric. That join key is what lets a platform team correlate Claude Code telemetry with Coder's audit log and exectrace on `workspace_id` without each template author having to remember to wire it. A generic env block can't synthesize that, the template author would have to write the interpolation themselves and most won't.

Happy to add a generic `env` map alongside if that's useful for other vars, but I'd argue `telemetry` should stay as a first-class input for the resource-attribute injection alone.

Will rebase onto the post-#861 main and bump to 5.1.0 shortly.

morganl-ant:
Rebased onto post-#861 main and bumped to 5.1.0.

registry/coder/modules/claude-code/main.tf
```diff
@@ -163,6 +175,55 @@ resource "coder_env" "anthropic_base_url" {
...
+locals {
+  # Always inject Coder workspace identifiers so OTEL data can be joined with
+  # Coder's audit log / exectrace on workspace_id without per-template wiring.
+  otel_resource_attributes = merge(
+    var.telemetry.resource_attributes,
+    {
+      "coder.workspace_id"    = data.coder_workspace.me.id
+      "coder.workspace_name"  = data.coder_workspace.me.name
+      "coder.workspace_owner" = data.coder_workspace_owner.me.name
+      "coder.template_name"   = data.coder_workspace.me.template_name
```
matifali:
```suggestion
      "coder.workspace_id"             = data.coder_workspace.me.id
      "coder.workspace_name"           = data.coder_workspace.me.name
      "coder.template_name"            = data.coder_workspace.me.template_name
      "coder.template_version"         = data.coder_workspace.me.template_version
      "coder.access_url"               = data.coder_workspace.me.access_url
      "coder.workspace_owner_id"       = data.coder_workspace_owner.me.id
      "coder.workspace_owner_username" = data.coder_workspace_owner.me.name
```
A few more items that would help enrich telemetry.

matifali (approved):
Thanks for the contribution. A few suggestion to further enrich Coder specific telemetry, but this looks perfect to me.

DevelopmentCats (approved):
This looks perfect now! If we want to add the other telemetry fields, we can get this merged and then go through the other ones as well. @morganl-ant

Issue #284, opened 2025-08-03
claude code: use `--append-system-prompt` to set system prompt instead of CLAUDE.md


bpmct:
I think this will significantly change its behavior and effectiveness

Majain004:
@DevelopmentCats PR ready for review

matifali:
Probably not needed now after https://github.com/coder/registry/pull/863

PR #863, opened 2026-04-22
feat(claude-code): add managed_settings input for policy delivery via /etc/claude-code

### LINKED ISSUE #818: claude-code: Add 'auto' and 'dontAsk' to permission_mode validation
## Problem

The `permission_mode` variable in the claude-code module only accepts `["", "default", "acceptEdits", "plan", "bypassPermissions"]`, but Claude Code CLI v2.1.85+ supports additional modes: `auto` and `dontAsk`.

```
claude --help
--permission-mode <mode>  (choices: "acceptEdits", "bypassPermissions", "default", "dontAsk", "plan", "auto")
```

### Why this matters

When using `bypassPermissions` mode in headless/automated setups (Coder Tasks), Claude Code v2.1.85 shows an interactive confirmation prompt:

```
WARNING: Claude Code running in Bypass Permissions mode.
Type 'yes' to accept:
```

This blocks indefinitely in non-interactive environments (agentapi PTY) because no one can type "yes". The `skipDangerousModePermissionPrompt` setting in `settings.json` is supposed to suppress this but doesn't work reliably in v2.1.85.

The `auto` mode is a better fit for headless agents — it uses a background safety classifier instead of showing scary interactive prompts, and it doesn't require the user to accept a bypass dialog.

### 1. Add `auto` and `dontAsk` to `permission_mode` validation

```hcl
validation {
  condition     = contains(["", "default", "acceptEdits", "plan", "bypassPermissions", "auto", "dontAsk"], var.permission_mode)
  error_message = "permission_mode must be one of: default, acceptEdits, plan, bypassPermissions, auto, dontAsk."
}
```

### 2. Don't hardcode `--dangerously-skip-permissions` for tasks when another mode is set

Currently in `start.sh`, tasks always get `--dangerously-skip-permissions` regardless of the configured `permission_mode`:

```bash
ARGS+=(--session-id "$TASK_SESSION_ID" --dangerously-skip-permissions)
```

When `permission_mode` is explicitly set to `auto` or `dontAsk`, this hardcoded flag should be omitted so the chosen mode is respected.

Suggested logic:
```bash
ARGS+=(--session-id "$TASK_SESSION_ID")
if [ -z "$ARG_PERMISSION_MODE" ] || [ "$ARG_PERMISSION_MODE" = "bypassPermissions" ]; then
  ARGS+=(--dangerously-skip-permissions)
fi
```

## Context

- Claude Code `auto` mode was released March 24, 2026
- Requires Team/Enterprise plan + Sonnet 4.6 or Opus 4.6
- No interactive prompts, uses a background classifier for safety
- See: https://www.anthropic.com/engineering/claude-code-auto-mode

morganl-ant:
## Problem

The module configures Claude Code's permission posture by reaching around the permission system rather than through it:

- `scripts/install.sh` writes `bypassPermissionsModeAccepted`, `autoModeAccepted`, and `primaryApiKey` directly into the user-writable `~/.claude.json`. Any process in the workspace can read the API key or flip the acceptance flags back.
- `scripts/start.sh` adds `--dangerously-skip-permissions` to every task launch, even when the template author set an explicit `permission_mode`. The README has to carry a security warning telling people the module bypasses permission checks.
- `permission_mode`, `allowed_tools`, and `disallowed_tools` each plumb through a different ad-hoc path (CLI flag, `coder` subcommand) instead of a single policy surface.

## Change

Add a `managed_settings` input that renders to `/etc/claude-code/managed-settings.d/10-coder.json`. Claude Code reads that drop-in directory at startup with the highest configuration precedence (above `~/.claude/settings.json` and project settings), so template authors get an admin-controlled policy file that users inside the workspace cannot override. The mechanism is a local file read with no API call, so it works identically for the Anthropic API, AWS Bedrock, Google Vertex AI, and AI Bridge / AI Gateway.

```hcl
managed_settings = {
  permissions = {
    defaultMode                  = "acceptEdits"
    disableBypassPermissionsMode = "disable"
    deny                         = ["Bash(curl:*)", "WebFetch"]
  }
}
```

Supporting changes:

- `install.sh` writes the policy file (root-owned, 0644) and stops writing `bypassPermissionsModeAccepted`, `autoModeAccepted`, and `primaryApiKey` into `~/.claude.json`. The API key is already exported via `coder_env` as `CLAUDE_API_KEY`; duplicating it on disk is unnecessary. `hasCompletedOnboarding` stays because there is no env-var alternative for it.
- `start.sh` only adds `--dangerously-skip-permissions` for tasks when no explicit `permission_mode` is set (same fix as #846; included here so this PR is self-contained, happy to drop if #846 lands first).
- `permission_mode`, `allowed_tools`, and `disallowed_tools` are marked deprecated and shimmed into `managed_settings.permissions` for one release when `managed_settings` is not provided.
- README security warning rewritten to point at the policy mechanism instead of telling people the module is unsafe by design.

## Relationship to #861

#861 strips this module to install-and-configure and removes `permission_mode` / `allowed_tools` / `disallowed_tools` outright. `managed_settings` is the natural replacement for those: it is install-time (survives the `start.sh` removal), it covers everything the dropped variables did plus `hooks`, `env`, `model`, `apiKeyHelper`, and the rest of the settings schema, and it does not require the module to know anything about how Claude is launched. If #861 lands first I will rebase this on top and drop the deprecation shim and the `start.sh` hunk.

## Validation

- `terraform fmt` / `terraform validate` clean
- New tests: `claude-managed-settings-written`, `claude-managed-settings-legacy-shim`, `claude-no-policy-keys-in-claudejson`, plus an assertion in `claude-auto-permission-mode` that `--dangerously-skip-permissions` is absent when a mode is set
- Manually verified `/etc/claude-code/managed-settings.d/*.json` precedence in the Claude Code CLI source

Closes #818. Relates to #284, #846, #861.

Disclosure: I work at Anthropic on the Claude Code team. Happy to adjust scope or split this further if that is easier to review.


matifali:
This would need rebasing after #861 is merged. Thanks for your contribution.

morganl-ant:
Rebased onto post-#861 main. Scope is now smaller: just the `managed_settings` input writing `/etc/claude-code/managed-settings.d/10-coder.json`. The deprecation shim and `~/.claude.json` key removal from the original PR are no longer needed since #861 dropped those vars and writes.

registry/coder/modules/claude-code/scripts/install.sh.tftpl
```diff
@@ -158,8 +186,6 @@ function configure_standalone_mode() {
     echo "Updating existing Claude configuration at $${claude_config}"
 
     jq '.autoUpdaterStatus = "disabled" |
-        .autoModeAccepted = true |
-        .bypassPermissionsModeAccepted = true |
```
matifali, line 162: Why did we remove this? AFAIK, these were needed for skipping the welcome wizard.

matifali (approved):
I like the flexibility this PR adds. I think this supercedes #866 and covers the use case.

registry/coder/modules/claude-code/scripts/install.sh.tftpl
DevelopmentCats: I think the reason they are removing them is because you are able to set these same behaviors through managed settings. Which is probably better than us pre-seeding it for all users in this hacky unobservable way.

Although I think we should probably test this a bit more before committing since we want to make sure this won't introduce too big of a behavioral shift in the module for users who have this running in a task workflow.

registry/coder/modules/claude-code/main.tftest.hcl
```diff
@@ -283,3 +283,47 @@ run "test_workdir_optional" {
...
+  command = plan
+
+  variables {
+    agent_id = "test-agent-managed-settings"
+    workdir  = "/home/coder/project"
+    managed_settings = {
+      permissions = {
+        defaultMode                  = "acceptEdits"
+        disableBypassPermissionsMode = "disable"
+        deny                         = ["Bash(rm -rf*)"]
```
DevelopmentCats, line 297: @matifali for reference this is how you would pass similar behavior through managed settings rather than pre-seeding it.

So essentially someone could set `defaultMode=auto` and same with the rest for claude-code running for tasks and this would work over even the `settings.json`

DevelopmentCats:
Everything looks good here for me @morganl-ant I will go ahead and test this and approve it from my end once done.

registry/coder/modules/claude-code/scripts/install.sh.tftpl
matifali: This serves the standalone mode, which we need to set anyway when a user provides a `workdir`. I don't see how that would work if we remove this. I am all good if we can do it more systematically. But if a user starts Claude via `coder_script` in a specific `workdir`, they should not be greeted with a config wizard.

DevelopmentCats: You know I think you are right here. These are mostly just for the dialogue boxes unless they get suppressed now somehow.

DevelopmentCats:
@morganl-ant @matifali 

I have tested this and it works properly with agentapi/tasks provided you do the following in the new module

```
  claude_managed_settings = {
    permissions = {
      defaultMode = "bypassPermissions"
      allow       = ["*"]
      deny        = []
    }
    # Skip the first-run theme picker / onboarding prompts.
    theme                  = "dark"
    hasCompletedOnboarding = true
  }

```

matifali:
@DevelopmentCats, I guess we can safely remove (deprecate) the `workdir` input then.

DevelopmentCats:
> @DevelopmentCats, I guess we can safely remove (deprecate) the `workdir` input then.

Yeah we definitely can if the only thing that was holding us back was project level configs.

DevelopmentCats:
@matifali I think we are good to merge this. If we decide later that we want to add back the part where we auto fill the config info the managed settings to bypass the prompts by default, but its probably fine to let the users or administrator decide what they need for the managed settings in the module to make it perform how they'd like. Especially with the move to agents.

matifali:
@DevelopmentCats Can you test the `coder_app` example use case? If it can start Claude without user needing to do anything I am fine merging this.

DevelopmentCats:
> @DevelopmentCats Can you test the `coder_app` example use case? If it can start Claude without user needing to do anything I am fine merging this.

Yeah I will test and provide a picture

matifali:
@DevelopmentCats can also update example with required settings (if needed)

DevelopmentCats:
@morganl-ant Is there a way to suppress either of these Claude-Code CLI alerts through a setting that can be done through `managed_settings`?

image

image

morganl-ant:
Short answer: no — those two screens can't be suppressed via `managed_settings`.

The theme picker and the security-notes screen are one first-run onboarding flow, gated by `hasCompletedOnboarding` in **`~/.claude.json`** — the state file, not the settings schema. Pre-setting `theme` in managed settings doesn't skip the picker; both screens show or skip together.

So the split is:
- **`managed_settings` (this PR)** → admin-enforced *policy* (`permissions`, `cleanupPeriodDays`, `hooks`, `env`, …)
- **`~/.claude.json` seed** → first-run *state* (`hasCompletedOnboarding`, `hasTrustDialogAccepted`)

The module should keep writing `{"hasCompletedOnboarding": true}` to `~/.claude.json` alongside the new `managed_settings` input — they're complementary. This PR doesn't regress that, so good to merge from my side.

(Onboarding is also skipped automatically in non-interactive runs — `claude -p` or when `CI` is set — but setting `CI` workspace-wide would confuse every other tool, so I wouldn't lean on that here. The lack of a dedicated "skip onboarding" key for provisioned environments is a fair gap on our side; I'll track it.)

DevelopmentCats (approved):

PR #865, opened 2026-04-22
feat(claude-code): add api_key_helper for short-lived credentials; mark anthropic_api_key sensitive

morganl-ant:
## Problem

The module currently treats the Anthropic API key as a static, long-lived secret:

- `claude_api_key` is not marked `sensitive`, so it can appear in plain text in Terraform plan output and CI logs (only `claude_code_oauth_token` had the flag).
- `install.sh` writes the key to `~/.claude.json` as `primaryApiKey`. The CLI already reads the key from the `ANTHROPIC_API_KEY` / `CLAUDE_API_KEY` environment variable that this module sets via `coder_env`, so the disk write is redundant and leaves a credential at rest in a user-readable file.
- There is no first-class way to use short-lived credentials. Customers who source keys from Vault, AWS Secrets Manager, or cloud IAM today have to wrap the module with a `pre_install_script` and still end up with a static value in TF state.

## Changes

- New `api_key_helper` input (`{ script, ttl_ms }`). The script is written to `~/.claude/coder-api-key-helper.sh` (mode 0700) and registered via the Claude Code [`apiKeyHelper`](https://docs.anthropic.com/en/docs/claude-code/settings#available-settings) setting at `/etc/claude-code/managed-settings.d/20-coder-apikeyhelper.json`. Claude invokes the helper whenever it needs a key and caches the result for `ttl_ms` (default 5 minutes), so the credential never lands in TF state, the agent environment, or `~/.claude.json`.
- `coder_env` for `CLAUDE_CODE_API_KEY_HELPER_TTL_MS` when `api_key_helper` is set.
- `claude_api_key` now has `sensitive = true` and an updated description.
- `install.sh` no longer writes `primaryApiKey` to `~/.claude.json`. Onboarding-bypass keys (`hasCompletedOnboarding`, `hasTrustDialogAccepted`, etc.) are kept so the CLI still starts non-interactively. The early-return guard is widened so onboarding bypass also runs when `api_key_helper` is the auth source.
- Validation: `api_key_helper` is mutually exclusive with `claude_api_key`, `claude_code_oauth_token`, and `enable_aibridge`.
- README: new "Short-lived credentials via api_key_helper" section with Vault and AWS Secrets Manager examples.

This is install-time configuration only, so it composes cleanly with the direction in #861.

## Validation

- `terraform fmt` clean
- `terraform validate` clean
- `terraform test`: 23/23 pass (4 new: `test_api_key_helper`, `test_api_key_helper_default_ttl`, `test_api_key_helper_validation_with_api_key`, `test_api_key_helper_validation_with_aibridge`)
- `bun test`: 2 new tests pass (`api-key-helper`, `claude-api-key-not-written-to-claude-json`); existing suite untouched
- `shellcheck --severity=warning scripts/install.sh` clean

## Disclosure

I work at Anthropic on the Claude Code team. This change is part of a small batch of upstream contributions based on what we have seen enterprise customers ask for; happy to adjust scope or split further if that helps review.

### DIFF
--- a/registry/coder/modules/claude-code/main.tf
+++ b/registry/coder/modules/claude-code/main.tf
@@ -62,7 +62,8 @@ variable "disable_autoupdater" {
 
 variable "anthropic_api_key" {
   type        = string
-  description = "API key passed to Claude Code via the ANTHROPIC_API_KEY env var."
+  description = "API key passed to Claude Code via the ANTHROPIC_API_KEY env var. Prefer api_key_helper for short-lived credentials."
+  sensitive   = true
   default     = ""
 }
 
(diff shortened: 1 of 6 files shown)


morganl-ant:
Rebased onto post-#861 main. Var rename `claude_api_key` -> `anthropic_api_key` picked up; the `primaryApiKey` removal from the original PR is no longer needed since #861's install template doesn't write it.

registry/coder/modules/claude-code/scripts/install.sh.tftpl
```diff
@@ -144,11 +146,40 @@ function setup_claude_configurations() {
...
+  fi
+
+  echo "Configuring api_key_helper for short-lived credentials..."
+
+  mkdir -p "$HOME/.claude"
+  local helper_path="$HOME/.claude/coder-api-key-helper.sh"
+  printf '%s' "$${ARG_API_KEY_HELPER_SCRIPT}" > "$${helper_path}"
+  chmod 0700 "$${helper_path}"
+
+  local dropin_dir="/etc/claude-code/managed-settings.d"
```
matifali, line 189: I wonder how this ties into #863

registry/coder/modules/claude-code/main.tf
```diff
@@ -118,6 +119,25 @@ variable "enable_ai_gateway" {
   }
 }
 
+variable "api_key_helper" {
+  type = object({
+    script = string
+    ttl_ms = optional(number, 300000)
+  })
+  description = "Script that prints an Anthropic API key to stdout. Written to ~/.claude/coder-api-key-helper.sh and registered via the apiKeyHelper setting in /etc/claude-code/managed-settings.d/. Use for short-lived credentials from Vault, AWS Secrets Manager, cloud IAM, etc. ttl_ms is how long Claude Code caches each key (default 5 minutes)."
+  default     = null
```
matifali, line 146: Do we still need as explicit if we can do the same via #863

morganl-ant:
Will rebase once #863 merges so we're not chasing conflicts twice. Shout if you'd prefer this land first.

DevelopmentCats:
> Will rebase once #863 merges so we're not chasing conflicts twice. Shout if you'd prefer this land first.

#863 Is merged we should be good to rebase and move forwards with this one at well @morganl-ant

morganl-ant:
Rebased onto main post-#863.

DevelopmentCats (approved):

PR #866, opened 2026-04-22
feat(claude-code): add transcript_retention_days and Stop-hook idle sentinel

### LINKED ISSUE #726: claude-code: stale session state causes "Session ID already in use" after workspace restart
## Problem

When a workspace is terminated unexpectedly (pod killed, node drained, etc.), Claude's session state files persist on storage. On restart, the `start.sh` script detects the existing session file via `task_session_exists` + `is_valid_session` and uses `--resume`, but Claude may reject the session as conflicting if other stale metadata files are in an inconsistent state.

Additionally, if the session file exists but is in a state that passes validation yet Claude considers it "in use" (e.g. incomplete shutdown), there is no recovery path — the script has no mechanism to clean up stale session metadata and retry with a fresh session.

## Root Cause

Claude CLI creates multiple files to track session state:
- `~/.claude/projects/<dir>/<session-id>.jsonl` — session transcript
- `~/.claude/todos/<session-id>*.json` — todo state
- `~/.claude/debug/<session-id>.txt` — debug logs
- `~/.claude/session-env/<session-id>` — environment state

The current `start.sh` script only checks for and validates the `.jsonl` session transcript file ([`task_session_exists`](https://github.com/coder/registry/blob/main/registry/coder/modules/claude-code/scripts/start.sh#L97-L107) and [`is_valid_session`](https://github.com/coder/registry/blob/main/registry/coder/modules/claude-code/scripts/start.sh#L109-L147)). It does not:

1. Clean up stale metadata files in `~/.claude/todos/`, `~/.claude/debug/`, or `~/.claude/session-env/` on startup
2. Handle the case where `--resume` fails (no fallback to starting a fresh session)
3. Use any form of process-level lock (e.g. `flock`) to distinguish between an active session and leftover files from a dead process

## Suggested Fix

1. **Add stale session cleanup on startup**: Before attempting to resume, remove orphaned metadata files (`todos`, `debug`, `session-env`) for the hardcoded `TASK_SESSION_ID` when no Claude process is running.
2. **Add fallback logic**: If `--resume` fails, fall back to removing the stale session file and starting fresh with `--session-id`.
3. **Consider `flock`-based locking**: Use file locking to detect whether a session is genuinely active vs. leftover from a killed process.

## Workaround

Manually remove stale session metadata before starting:
```bash
rm -f ~/.claude/projects/*/<session-id>.jsonl
rm -f ~/.claude/todos/<session-id>*.json
rm -f ~/.claude/debug/<session-id>.txt
rm -f ~/.claude/session-env/<session-id>
```

---

Created on behalf of @matifali

morganl-ant:
## Problem

Two install-time lifecycle gaps in the v5 module:

- Session JSONL transcripts under `~/.claude/projects/` are never pruned by the module, so long-lived workspaces accumulate unbounded transcript files unless the user knows to set `cleanupPeriodDays` themselves.
- There is no machine-readable signal that Claude has gone idle, which makes it hard for templates to drive Coder's workspace autostop or activity tracking off agent state.

## Changes

- New `transcript_retention_days` input (number, default null, validated >= 1). When set, the module writes `cleanupPeriodDays` to a managed-settings drop-in so Claude Code prunes old transcripts automatically. When unset, Claude Code's built-in 30-day default applies.
- The same drop-in registers a `Stop` hook that touches `~/.coder-modules/coder/claude-code/last-stop` whenever Claude finishes a turn. Templates can read that file's mtime to drive autostop or activity bumping.
- Both are written to `/etc/claude-code/managed-settings.d/30-coder-lifecycle.json`. This is the local managed-settings file mechanism, read by the Claude CLI regardless of inference backend (Anthropic API, Bedrock, Vertex, AI Gateway). If `/etc/claude-code` is not writable and `sudo` is unavailable, install logs a warning and continues.
- Tests: `lifecycle-settings-written` (retention set) and `lifecycle-settings-default-retention` (Stop hook present, no `cleanupPeriodDays`).
- README: new "Session lifecycle" section; version bumped to 5.1.0.

## Relationship to #861

The original revision of this PR also replaced the hardcoded `TASK_SESSION_ID` in `start.sh` with a workspace-derived UUIDv5 (#726). #861 removed `start.sh` and module-managed sessions entirely, which resolves #726 directly: there is no longer a hardcoded session ID because the module no longer starts Claude. This PR is now reduced to the install-time pieces above.

## Validation

- `terraform fmt` clean
- `terraform validate` clean
- `terraform test` 14/14 pass
- `bun test` 14/14 pass locally (one pre-existing docker-cleanup race in `claude-mcp-config` is flaky in full-suite runs but passes in isolation; unrelated to this change)

---
Disclosure: I work on Claude Code at Anthropic. This is one of a small set of contributions we are proposing to the `claude-code` module; happy to discuss the overall direction in whichever venue the maintainers prefer.

### DIFF
--- a/registry/coder/modules/claude-code/README.md
+++ b/registry/coder/modules/claude-code/README.md
@@ -36,6 +36,25 @@ module "claude-code" {
 
 By default, Claude Code automatically resumes existing conversations when your workspace restarts. Sessions are tracked per workspace directory, so conversations continue where you left off. If no session exists (first start), your `ai_prompt` will run normally. To disable this behavior and always start fresh, set `continue = false`
 
+### Session lifecycle
+
+When task reporting is enabled the module pins Claude Code to a session ID derived from `data.coder_workspace.me.id` (UUIDv5). This keeps the conversation stable across restarts of the same workspace while remaining unique per workspace, avoiding the "Session ID already in use" error that can occur when home directories are templated or shared.
+
+The module also writes a managed settings drop-in at `/etc/claude-code/managed-settings.d/30-coder-lifecycle.json` that:
+
+- registers a `Stop` hook which touches `~/.claude-module/last-stop` whenever Claude finishes a turn, so template authors can wire workspace autostop or activity tracking off that file's modification time
+- sets `cleanupPeriodDays` when `transcript_retention_days` is provided, so session JSONL transcripts are pruned automatically
+
+```tf
+module "claude-code" {
+  source                    = "registry.coder.com/coder/claude-code/coder"
+  version                   = "4.9.2"
+  agent_id                  = coder_agent.main.id
+  workdir                   = "/home/coder/project"
+  transcript_retention_days = 7
+}
+```
+
 ## State Persistence
 
 AgentAPI can save and restore its conversation state to disk across workspace restarts. This complements `continue` (which resumes the Claude CLI session) by also preserving the AgentAPI-level context. Enabled by default, requires agentapi >= v0.12.0 (older versions skip it with a warning).
--- a/registry/coder/modules/claude-code/README.md
+++ b/registry/coder/modules/claude-code/README.md
@@ -13,7 +13,7 @@ Install and configure the [Claude Code](https://docs.anthropic.com/en/docs/agent
 ```tf
 module "claude-code" {
   source            = "registry.coder.com/coder/claude-code/coder"
-  version           = "5.0.0"
+  version           = "5.1.0"
   agent_id          = coder_agent.main.id
   anthropic_api_key = "xxxx-xxxxx-xxxx"
 }
@@ -47,7 +47,7 @@ locals {
 
 module "claude-code" {
   source            = "registry.coder.com/coder/claude-code/coder"
-  version           = "5.0.0"
+  version           = "5.1.0"
   agent_id          = coder_agent.main.id
   workdir           = local.claude_workdir
   anthropic_api_key = "xxxx-xxxxx-xxxx"
@@ -78,7 +78,7 @@ resource "coder_app" "claude" {
 ```tf
 module "claude-code" {
   source            = "registry.coder.com/coder/claude-code/coder"
-  version           = "5.0.0"
+  version           = "5.1.0"
   agent_id          = coder_agent.main.id
   workdir           = "/home/coder/project"
   enable_ai_gateway = true
@@ -102,7 +102,7 @@ This example shows version pinning, a pre-installed binary path, a custom model,
 ```tf
 module "claude-code" {
   source   = "registry.coder.com/coder/claude-code/coder"
-  version  = "5.0.0"
+  version  = "5.1.0"
   agent_id = coder_agent.main.id
   workdir  = "/home/coder/project"
 
@@ -166,7 +166,7 @@ Downstream `coder_script` resources can wait for this module's install pipeline 
 ```tf
 module "claude-code" {
   source            = "registry.coder.com/coder/claude-code/coder"
-  version           = "5.0.0"
+  version           = "5.1.0"
   agent_id          = coder_agent.main.id
   workdir           = "/home/coder/project"
   anthropic_api_key = "xxxx-xxxxx-xxxx"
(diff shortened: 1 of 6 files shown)


matifali (approved):
@morganl-ant thanks for your contribution. I think we can merge this to `main` and release a new patch version to unblock your users. Bit for other PRs, let's rebase them once #861 (likely today) lands.

matifali:
Also, you would need to bump the patch version in README.md examples and fix the failing tests.

morganl-ant:
Thanks, both addressed in 11925ea:

- README examples bumped 4.9.2 -> 4.9.3.
- Test failures were two separate bugs: (1) the `deriveTaskSessionId` helper tried to grep the workspace ID out of `script.sh`, but the install/start scripts are base64-encoded inside the agentapi wrapper so the regex never matched and the helper silently returned the legacy constant. Fixed by pinning `CODER_WORKSPACE_ID` in the test env so the derived session ID is deterministic. (2) `configure_lifecycle_settings` was called after `report_tasks` in install.sh; `report_tasks` shells out to the `coder` CLI which is not in the test container, and `set -e` aborted before the lifecycle file was written. Swapped the order.

Full suite is 23/23 locally now.

morganl-ant:
Rebased onto post-#861 main. The session-ID derivation half of the original PR fell away with #861's removal of `start.sh` (which also resolves #726 directly). What remains is the install-time `transcript_retention_days` -> `cleanupPeriodDays` mapping and the Stop-hook idle sentinel. Title and description updated to match.

registry/coder/modules/claude-code/scripts/install.sh.tftpl
```diff
@@ -144,6 +146,36 @@ function setup_claude_configurations() {
 
 }
 
+function configure_lifecycle_settings() {
+  local dropin_dir="/etc/claude-code/managed-settings.d"
+  local target="$${dropin_dir}/30-coder-lifecycle.json"
+  # Bake the absolute path at install time so the hook does not depend on
+  # $HOME being set identically when Claude Code executes it.
+  local sentinel="$HOME/.coder-modules/coder/claude-code/last-stop"
```
matifali, line 154: This should depend on `module_dir_name` input. Although we have a very strict regex there, but hardcoding this would be less flexible for any future changes.

registry/coder/modules/claude-code/README.md
```diff
@@ -189,6 +189,26 @@ resource "coder_script" "post_claude" {
 }
 ```
 
+### Session lifecycle
+
+The module writes a small managed-settings drop-in to `/etc/claude-code/managed-settings.d/30-coder-lifecycle.json` that:
+
+- registers a `Stop` hook which touches `~/.coder-modules/coder/claude-code/last-stop` whenever Claude finishes a turn. Templates can read that file's modification time to drive workspace autostop or activity tracking.
```
matifali, line 196: Can we elaborate a bit more on how templates can use this for activity tracking?

registry/coder/modules/claude-code/README.md
```diff
@@ -189,6 +189,26 @@ resource "coder_script" "post_claude" {
...
+  source                    = "registry.coder.com/coder/claude-code/coder"
+  version                   = "5.1.0"
+  agent_id                  = coder_agent.main.id
+  workdir                   = "/home/coder/project"
+  anthropic_api_key         = "xxxx-xxxxx-xxxx"
+  transcript_retention_days = 7
+}
+```
+
+The drop-in is a local file read by the Claude CLI at startup; it works with any inference backend (Anthropic API, Bedrock, Vertex, AI Gateway). If `/etc/claude-code` is not writable in the workspace image and `sudo` is unavailable, the install script logs a warning and skips the write.
```
matifali, line 210: Should we, in this case, write to the user's `$HOME`, e.g., `$HOME/.claude/settings.json`?

matifali (commented):
Claude Code is very fast-paced. And we see new features and settings being added every other week. Do you think we should have a more flexible generic way to configure it than adding new named module inputs for each new feature?

For example, https://code.claude.com/docs/en/settings shows many options that can be useful for Enterprises, but adding all of them as named modules does not seem scalable and maintainable to me. 

Curious about what you think about this?

Otherwise, I am happy to accept this. One reason for the larger refactor we did in v5.0.0 was to increase the flexibility of configuration so we do not limit users on how they want to use Claude code in a coder workspace. 

Thanks

DevelopmentCats:
@matifali 

I agree with you here. I think we can close this because this amounts to configuration that can now be done through `managed_settings` introduced in https://github.com/coder/registry/pull/863

matifali:
@DevelopmentCats I would wait for what @morganl-ant thinks. So maybe keep it open for now.

morganl-ant:
Agreed — once #863 lands, both halves of this become `managed_settings` entries:
```json
{
  "cleanupPeriodDays": 14,
  "hooks": {
    "Stop": [{ "hooks": [{ "type": "command", "command": "touch /tmp/claude-idle" }] }]
  }
}
```
I'd rather keep the module surface small and document this pattern in the README's `managed_settings` section than add two more top-level vars. Closing in favor of #863; I'll fold the example above into a follow-up docs commit.

PR #870, opened 2026-04-23
feat(coder-utils): nest logs under module_directory/logs

matifali:
## Summary

Move `pre_install.log`, `install.log`, `post_install.log`, and `start.log` from the flat `${module_directory}` to a `logs/` subdirectory. Keeps the scripts directory clean and separates executable artifacts from runtime output.

## Changes

- Compute `local.log_directory = "${var.module_directory}/logs"` and use it for every `*.log` path.
- Each `coder_script` `mkdir -p`s the `logs/` sub-path so the directory exists even when `post_install` or `start` is the first consumer after `install`.
- Update the `module_directory` description to call out the nested logs path.
- Update tee-regex tests to match the new `logs/` prefix.
- Add `test_logs_nested_under_module_directory` asserting the new paths and the `mkdir -p` in each script.
- Bump module version to `v1.2.0`.

## Breaking Changes

Consumers reading `${module_directory}/install.log` (and friends) directly must look under `${module_directory}/logs/` instead. The only in-repo consumer (`coder/claude-code`) is migrating in https://github.com/coder/registry/pull/861.

## Validation

- `terraform fmt -recursive` clean
- `terraform validate` clean
- `terraform test` → 15/15 pass
- `prettier --check` clean

### DIFF
--- a/registry/coder/modules/coder-utils/main.tf
+++ b/registry/coder/modules/coder-utils/main.tf
@@ -51,7 +51,7 @@ variable "agent_name" {
 
 variable "module_directory" {
   type        = string
-  description = "The module's working directory for scripts and logs."
+  description = "The module's working directory for the install/pre/post/start scripts this module writes. Logs land under a `logs/` subdirectory of this path."
 }
 
 variable "display_name_prefix" {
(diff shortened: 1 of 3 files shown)


registry/coder/modules/coder-utils/main.tf
```diff
@@ -178,6 +184,8 @@ resource "coder_script" "start_script" {
     set -o errexit
     set -o pipefail
 
+    mkdir -p ${local.log_directory}
+
```
35C4n0r: not needed, since install script is mandatory

registry/coder/modules/coder-utils/main.tf
```diff
@@ -156,6 +160,8 @@ resource "coder_script" "post_install_script" {
...
```
35C4n0r: not needed, since install script is mandatory

35C4n0r (commented):
minor nits, otherwise LGTM

35C4n0r (approved):

registry/coder/modules/coder-utils/main.tf
matifali: Addressed in 70a0093. `start` sync-depends on `install` (and `post_install` when configured), so `logs/` is already present.

registry/coder/modules/coder-utils/main.tf
matifali: Addressed in 70a0093. `post_install` sync-depends on `install` (which mkdirs `logs/`), so the directory is already there by the time this script runs.

PR #874, opened 2026-04-24
refactor(registry/coder/modules/coder-utils)!: derive names from module_directory

matifali:
## Summary

Derives `coder-utils` script names from `module_directory` instead of a separate `agent_name` input. The `module_directory` already encodes both the namespace and the module name, so carrying both is redundant and error-prone. Callers like `claude-code` no longer need to pass `agent_name`.

Scripts this module materializes lose the `${agent_name}-utils-` prefix because `module_directory` already namespaces them per-caller.

We will address multiple instances of coder-utils per caller in a future iteration if needed.

## Versioning Note

Previous tags (`v1.0.0` through `v1.3.0`) have been deleted because no published module ever consumed them — the module was effectively unreleased. This PR ships the first real public version as **`v0.0.1`**, treating it as a fresh start rather than a breaking bump from a version that was never in production use.

## Changes

- Remove `agent_name` variable.
- Derive `caller_name = "${namespace}-${module_name}"` from `module_directory`.
- Validate `module_directory` matches `$HOME/.coder-modules/<namespace>/<module-name>`.
- Rename script files on disk from `${agent_name}-utils-<phase>.sh` to plain `<phase>.sh`.
- Add a TS test for the `module_directory` validation.
- Ship as `v0.0.1` (first published version; all prior tags removed).

## Breaking Changes

| Before | After |
|---|---|
| `agent_name = "myagent"` | removed (derived from `module_directory`) |
| `module_directory = ".my-module"` | `module_directory = "$HOME/.coder-modules/<ns>/<name>"` (validated) |
| Script files `${agent_name}-utils-install.sh` | `install.sh` |
| Script sync names `${agent_name}-install_script` | `${namespace}-${module_name}-install_script` |

No callers were depending on the old format (prior tags were unpublished).

## Validation

- `terraform fmt -recursive` clean
- `terraform validate` clean
- `terraform test` → 17/17 pass
- `bun test registry/coder/modules/coder-utils` → 5/5 pass
- `prettier --check` clean

## Consumer

coder/registry#861 (`claude-code`) consumes this and is currently pinned to the commit SHA until this merges and ships as `v0.0.1`.

### DIFF
--- a/registry/coder/modules/coder-utils/README.md
+++ b/registry/coder/modules/coder-utils/README.md
@@ -13,18 +13,13 @@ tags: [internal, library]
 
 The Coder Utils module is a building block for modules that need to run multiple scripts in a specific order. It uses `coder exp sync` for dependency management and is designed for orchestrating pre-install, install, post-install, and start scripts.
 
-> [!NOTE]
->
-> - The `agent_name` should be the same as that of the agentapi module's `agent_name` if used together.
-
 ```tf
 module "coder_utils" {
   source  = "registry.coder.com/coder/coder-utils/coder"
-  version = "1.3.0"
+  version = "0.0.1"
 
   agent_id         = coder_agent.main.id
-  agent_name       = "myagent"
-  module_directory = ".my-module"
+  module_directory = "$HOME/.coder-modules/coder/claude-code"
 
   pre_install_script = <<-EOT
     #!/bin/bash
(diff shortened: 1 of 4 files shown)


35C4n0r:
@matifali If we merge this:
- Users won't have the freedom to write outside of the defined regex, even when they are building modules for their personal use.
- we need to make sure to mention we cannot have 2 instances of the `coder-utils` module write to same module_directory, even when both of them are instantiated in the same parent module, they'll need to choose 2 different paths.
- Every parent module that will use this module, will also need to implement that regex requirement, for example we'll need to copy the validation block to claude-code, gemini, copilot, codex etc., imo it wouldn't be a good UX for users considering we allow them to specify the module directory but enforce path restrictions.

matifali:
> Every parent module that will use this module, will also need to implement that regex requirement, for example we'll need to copy the validation block to claude-code, gemini, copilot, codex etc., imo it wouldn't be a good UX for users considering we allow them to specify the module directory but enforce path restrictions.

This is not required. terraform will fail even if the nested validation fails

> Users won't have the freedom to write outside of the defined regex, even when they are building modules for their personal use.

That is fine. Being restrictive is better than permissive. We can relax it later if needed.

> we need to make sure to mention we cannot have 2 instances of the coder-utils module write to same module_directory, even when both of them are instantiated in the same parent module, they'll need to choose 2 different paths.

I think it is ok. We are the only consumer of this module today. We can add support for multiple instances when needed.

35C4n0r (approved):

PR #887, opened 2026-05-06
feat(git-clone): add pre_clone_script parameter

ikkz:
## Summary

Add `pre_clone_script` parameter to the git-clone module, allowing users to run custom scripts before cloning a repository.

## Use Case

This solves SSH host key verification issues (e.g., "Host key verification failed") by enabling users to configure SSH settings before the clone operation, such as adding known hosts or setting `StrictHostKeyChecking no`.

```tf
module "git-clone" {
  count            = data.coder_workspace.me.start_count
  source           = "registry.coder.com/coder/git-clone/coder"
  version          = "1.3.0"
  agent_id         = coder_agent.example.id
  url              = "git@github.com:org/repo.git"
  pre_clone_script = <<-EOT
    #!/bin/bash
    mkdir -p ~/.ssh
    echo -e "Host github.com\n    StrictHostKeyChecking no\n" > ~/.ssh/config
    chmod 600 ~/.ssh/config
  EOT
}
```

Ref: https://discord.com/channels/747933592273027093/1447777180695396452/1447777180695396452

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v1.3.0`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

### DIFF
--- a/registry/coder/modules/git-clone/README.md
+++ b/registry/coder/modules/git-clone/README.md
@@ -14,7 +14,7 @@ This module allows you to automatically clone a repository by URL and skip if it
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "1.2.3"
+  version  = "1.3.0"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
 }
@@ -28,7 +28,7 @@ module "git-clone" {
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "1.2.3"
+  version  = "1.3.0"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
   base_dir = "~/projects/coder"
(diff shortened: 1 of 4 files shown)


DevelopmentCats (approved):
LGTM

DevelopmentCats:
> ### 💡 Codex Review
> https://github.com/coder/registry/blob/e24e6610ea3620052f669247b191394c138d67d2/registry/coder/modules/git-clone/run.sh#L43
> 
> 
> **P1 Badge Propagate pre-clone script failures**
> Make the pre-clone hook fail fast when it returns a non-zero status. In `run.sh`, `PRE_CLONE_TMP` is executed but its exit code is never checked, and the script does not use `set -e`, so a failing `pre_clone_script` (for example, a prerequisite check that calls `exit 1`) is silently ignored and cloning continues. This breaks the advertised “validate prerequisites before cloning” behavior and can leave workspaces starting with unmet preconditions.
> 
> ℹ️ About Codex in GitHub

This pattern exists in post clone, as well but the overall difference is here is that we don't `set -euo pipefail` at the top of the run script like we do in dot files and other modules but this can be addressed in another PR.

PR #890, opened 2026-05-14
refactor(registry/coder-labs/modules/gemini): remove agentapi, tasks, start script

Harsh9485:
## Description

Major refactor of the coder-labs/gemini module to mirror the coder/claude-code v5 changes from https://github.com/coder/registry/pull/861.

##Changes

* Replace module "agentapi" with module "coder_utils" (registry.coder.com/coder/coder-utils/coder v0.0.1)
* Replace scripts/install.sh with scripts/install.sh.tftpl (Terraform templatefile)
* Delete scripts/start.sh
* Module dir changed from .gemini-module to .coder-modules/coder-labs/gemini 

### Removed variables
All AgentAPI pass-throughs, and start-script-only variables:
`order`, `group`, `report_tasks`, `cli_app`, `web_app_display_name`, `cli_app_display_name`, `install_agentapi`, `agentapi_version`, `ai_prompt`, `system_prompt`

I removed the `configure_mcp` function because `coder exp mcp configure gemini` does not support Gemini yet, and it breaks the script.

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally


DevelopmentCats:
@Harsh9485 Testing this in my deployment as well.

DevelopmentCats:
@Harsh9485 Gemini configures and installs properly. I would take a look at the review comments above, and I will also give this a final pass myself so that way once the other comments are addressed we will be good to merge

registry/coder-labs/modules/gemini/README.md
```diff
@@ -134,24 +110,25 @@ For enterprise users who prefer Google's Vertex AI platform:
...
+  workdir        = "/home/coder/project"
   use_vertexai   = true
 }
 ```
 
 ## Troubleshooting
 
-- If Gemini CLI is not found, ensure your API key is valid (`install_gemini` defaults to `true`)
-- Check logs in `~/.gemini-module/` for install/start output
-- Use the `gemini_api_key` variable to avoid requiring Google sign-in
```
DevelopmentCats, line 149: This is probably still an accurate point. so why remove it?

registry/coder-labs/modules/gemini/main.tf
```diff
@@ -51,99 +51,69 @@ variable "install_gemini" {
...
-  description = "Task prompt for automated Gemini execution"
-  default     = ""
-}
-
 variable "additional_extensions" {
   type        = string
-  description = "Additional extensions configuration in json format to append to the config."
+  description = "Additional extensions configuration in JSON format to append to the config."
   default     = null
 }
```
DevelopmentCats: What do you think about adding a separate mcp variable like claude-code where the user can add mcp servers seperately. I know gemini can add mcp servers through cli that way you wouldn't have to worry about ruining the gemini `settings.json`

Harsh9485: Do you mean removing the `additional_extensions` variable and letting users add MCP servers when `coder_app.command` runs? If so, that sounds like a good idea.

Users are already familiar with adding MCP servers in Gemini, so it could also improve the UX. Adding JSON directly into Terraform is a bit risky for both users and maintainers.

That said, the current feature is still useful when a team wants to use the same MCP servers and share a common configuration for quick setup. Otherwise, adding every MCP server manually for each workspace could become annoying.

registry/coder-labs/modules/gemini/README.md
Harsh9485: They were not removed, just replaced with the new log file location.

In my opinion, if someone is using Coder, they’re probably not such a beginner that they need the first line saying “ensure your API key is valid.” But that’s just my opinion. If you want, I can add that point back.

The second point was only about the log file location, so I think it was fine to remove it. However, the third point is important—that was my mistake, and I’ll add it back.

registry/coder-labs/modules/gemini/main.tf
DevelopmentCats: I was more or less thinking that after the install we could almost run the `gemini mcp add` commands to register any mcp servers without having to include them directly in the `settings.json` It would be best to probably do this after the settings.json is written as well.

Because right now its basically just appending the extensions onto the end of the default or provided gemini `settings.json`. Which is probably fine but since the cli does have the ability to add mcp's directly it just seemed like a good option to align it with claude-code.

registry/coder-labs/modules/gemini/README.md
DevelopmentCats: Its probably fine to leave it out It was just ultimately a NIT since it originally was there.

Although you are probably right its basically the first issue anyone should think of when dealing with agent cli's like this so its probably not a super useful callout.

registry/coder-labs/modules/gemini/main.tf
```diff
@@ -155,72 +125,40 @@ locals {
       "server"
     ],
     "env": {
-      "CODER_MCP_APP_STATUS_SLUG": "${local.app_slug}",
+      "CODER_MCP_APP_STATUS_SLUG": "gemini",
       "CODER_MCP_AI_AGENTAPI_URL": "http://localhost:3284"
```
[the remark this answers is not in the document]
Harsh9485: Oh, @matifali, I didn’t notice that. Is the `coder` binary already updated for this? I mean, when I remove it, `coder` doesn’t throw an error for a missing environment variable?

registry/coder-labs/modules/gemini/main.tf
matifali: post_install_script can be used for this purpose.

registry/coder-labs/modules/gemini/main.tf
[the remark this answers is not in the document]
matifali: All `CODER_MCP_*` environment variables should be removed. Check Claude Code v5.0 for reference.

registry/coder-labs/modules/gemini/main.tf
DevelopmentCats: I mean that is true. We just had the same convention in the claude-code module so I figured it was worth mentioning.

But no worries we can disregard

DevelopmentCats:
@Harsh9485 Once everything is addressed from Atif's review we should be good here.

Harsh9485:
Guys, I just wanted to let you know that I was involved in a bike accident, so I may be offline for a few days.

Harsh9485:
@matifali, I committed the basic suggestions, but I’m facing issues integrating the gemini mcp add feature like claude-code.

* The settings.json file is not created until I run the gemini command. Because of that, the mcp add command cannot create the settings.json file by itself.
* At first, I thought I made a mistake, but when I manually ran the mcp add command, it showed that the MCP was added. However, when I checked the Docker files, I only saw updated activity timestamps and no changes in the settings.json file or any file.

Screenshot 2026-05-20 204726
Screenshot 2026-05-20 205028

Harsh9485:
the new commit shows the current code logic.

matifali:
https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/

@Harsh9485 Looks like Gemini CLI is going away.

Harsh9485:
Now we have the Antigravity CLI to add.

matifali:
Closing this in favor of the new antigravity CLI module.

Issue #86, opened 2025-05-09
git-clone: Auto accept Host-Key


ReptoxX:
When trying to clone a git-repository through ssh it always fails because it's host-key is not confirmed / not trusted.
I would like to add the host-key before cloning so the git clone command would not fail. Or alternatively just auto-accept the hostkey.

moo-im-a-cow:
to auto accept the ssh key, you could make a script in the template that runs 
`ssh-keyscan github.com >> ~/.ssh/known_hosts`

Issue #60, opened 2024-07-04
Cloning with SSH URLs fail for new workspaces using git_clone module.


pachori-ujjwal:
### Description

The `git_clone` module fails to clone a repository on workspace startup if the workspace does not contain a `.ssh/known_hosts` with an entry of the git provider that is being used to clone with repository from.

The module gives a misleading error stating "Host key verification failed" and asks the user to add the keys to Github/Gitlab.

### Versions

git_clone module: 1.0.12
coder: v2.12.3+534d4ea

geiseri:
I think it might actually be related to the order of events.  When I create the desktop it states that I do not have access, but when I am in the session I can clone no issues.   I do see in the logs the checkout happen before the key scanning so I am betting that is it.

Eun:
I think the right solution would be to use ssh's `VerifyHostKeyDNS` option:

```terraform
module "git_clone" {
  count       = data.coder_parameter.git_repo.value != "" ? 1 : 0
  source      = "registry.coder.com/modules/git-clone/coder"
  version     = ">= 1.0.0"
  agent_id    = coder_agent.main.id
  url         = data.coder_parameter.git_repo.value
  base_dir    = dirname(local.project_dir)
  folder_name = basename(local.project_dir)
  branch_name = data.coder_parameter.git_branch.value

  pre_clone_script = <<-EOF
    mkdir -p ~/.ssh
    cat > ~/.ssh/config << 'SSH_CONFIG'
    Host *
        StrictHostKeyChecking accept-new
        VerifyHostKeyDNS yes
    SSH_CONFIG
    chmod 700 ~/.ssh
    chmod 600 ~/.ssh/config
  EOF
}
```

PR #891, opened 2026-05-15
fix(git-clone): propagate pre/post-clone script failures

ikkz:
## Description

Fix git-clone module to fail fast when `pre_clone_script` or `post_clone_script` returns a non-zero exit code. Previously, both scripts were executed but their exit codes were never checked — a failing pre-clone hook (e.g., a prerequisite check that calls `exit 1`) was silently ignored and cloning continued. This broke the advertised "validate prerequisites before cloning" behavior and could leave workspaces starting with unmet preconditions.

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v1.3.1`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

- https://github.com/coder/registry/pull/887#issuecomment-4413765491
- https://github.com/coder/registry/issues/60
- https://github.com/coder/registry/issues/86

### DIFF
--- a/registry/coder/modules/git-clone/README.md
+++ b/registry/coder/modules/git-clone/README.md
@@ -14,7 +14,7 @@ This module allows you to automatically clone a repository by URL and skip if it
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "1.3.0"
+  version  = "1.3.1"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
 }
@@ -28,7 +28,7 @@ module "git-clone" {
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "1.3.0"
+  version  = "1.3.1"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
   base_dir = "~/projects/coder"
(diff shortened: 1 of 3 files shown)


DevelopmentCats:
Hey, thanks for looking into this! The root cause here is actually that `run.sh` is missing `set -euo pipefail` at the top, which most of our newer modules already use (claude-code, aider, amazon-q, agent-firewall, dotfiles, etc.). Adding that one line would make the manual `if !` guards unnecessary since the script would automatically exit on any non-zero return, and it'd also catch other silent failures we're not handling today (like `git clone` itself failing, or the `base64 -d` pipe breaking).

@ikkz Would you be open to swapping the approach to just add `set -euo pipefail` at the top instead? Smaller diff, catches more cases, and brings this module in line with the rest of the registry.

ikkz:
@DevelopmentCats Updated as suggested.

Added `set -euo pipefail` to `run.sh` instead of manually checking the pre/post-clone script exit codes, and added regression tests for failing pre/post-clone hooks.

Also, once this PR is merged, when should we expect the new `git-clone` module version (`v1.3.1`) to be published?

DevelopmentCats (approved):

DevelopmentCats:
> @DevelopmentCats Updated as suggested.
> 
> Added `set -euo pipefail` to `run.sh` instead of manually checking the pre/post-clone script exit codes, and added regression tests for failing pre/post-clone hooks.
> 
> Also, once this PR is merged, when should we expect the new `git-clone` module version (`v1.3.1`) to be published?

Once we merge this PR I will immediately go and release this Module and it should be available a few minutes after that.

DevelopmentCats:
@ikkz  Should show up on the registry within the next 10 mins.

ikkz:
Thank you! Much appreciated.

PR #896, opened 2026-05-20
fix(coder-labs/modules/codex): prevent config.toml overwrite on restart

35C4n0r:
## Description

Replace the destructive overwrite of `~/.codex/config.toml` with a marker-block merge. The module owns a fenced region of the file; everything outside the markers is user-owned and preserved across restarts.

```toml
# >>> coder-managed: codex module >>>
preferred_auth_method = "apikey"
[mcp_servers.github]
command = "npx"
# <<< coder-managed: codex module <<<
```

Sample Template: https://dev.coder.com/templates/product/codex-pr-896/docs

## Type of Change
- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information
Path: `registry/coder-labs/modules/codex`
New version: v5.1.1

Breaking change: [ ] Yes [x] No

## Testing & Validation
- [x] Tests pass (`bun test`)
- [x] Changes tested locally

Closes: [REG-11](https://linear.app/codercom/issue/REG-11)

### DIFF
--- a/registry/coder-labs/modules/codex/main.test.ts
+++ b/registry/coder-labs/modules/codex/main.test.ts
@@ -425,6 +425,309 @@ describe("codex", async () => {
     expect(installLog).toContain("Installed Codex CLI");
   });
 
+  test("idempotent-defaults-preserve-user-edits", async () => {
+    const { id, scripts } = await setup();
+    await runScripts(id, scripts);
+
+    // User edits the config between restarts
+    await execContainer(id, [
+      "bash",
+      "-c",
+      `cat > /home/coder/.codex/config.toml << 'EOF'
+preferred_auth_method = "login"
+custom_user_key = "my_value"
+
+[projects."/home/coder/project"]
+trust_level = "trusted"
+EOF`,
+    ]);
+
+    // Second run: user edits must survive
+    await runScripts(id, scripts);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // User's overridden value preserved (not reset to "apikey")
+    expect(config).toMatch(/preferred_auth_method\s*=\s*['"]login['"]/);
+    // User's custom key preserved
+    expect(config).toMatch(/custom_user_key\s*=\s*['"]my_value['"]/);
+  });
+
+  test("idempotent-mcp-deep-merge", async () => {
+    const mcpConfig = [
+      "[mcp_servers.github]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-github"]',
+      'type = "stdio"',
+      "",
+      "[mcp_servers.filesystem]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-filesystem"]',
+      'type = "stdio"',
+    ].join("\n");
+    const { id, scripts } = await setup({
+      moduleVariables: { mcp: mcpConfig },
+    });
+    await runScripts(id, scripts);
+
+    // User customizes the github MCP server between restarts
+    await execContainer(id, [
+      "bash",
+      "-c",
+      [
+        "CONFIG=/home/coder/.codex/config.toml",
+        // Replace the github command the user has customized
+        "sed -i \"s/command = .npx./command = 'my-custom-npx'/\" $CONFIG",
+      ].join(" && "),
+    ]);
+
+    // Second run
+    await runScripts(id, scripts);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // User's customized github command preserved
+    expect(config).toMatch(/command\s*=\s*['"]my-custom-npx['"]/);
+    // filesystem server still present (not lost by shallow merge)
+    expect(config).toContain("mcp_servers");
+    expect(config).toContain("filesystem");
+  });
+
+  test("idempotent-base-config-preserves-user-edits", async () => {
+    const baseConfig = [
+      'sandbox_mode = "danger-full-access"',
+      'preferred_auth_method = "apikey"',
+    ].join("\n");
+    const { id, scripts } = await setup({
+      moduleVariables: { base_config_toml: baseConfig },
+    });
+    await runScripts(id, scripts);
+
+    // User changes sandbox_mode
+    await execContainer(id, [
+      "bash",
+      "-c",
+      "sed -i 's/danger-full-access/sandbox/' /home/coder/.codex/config.toml",
+    ]);
+
+    // Second run
+    await runScripts(id, scripts);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // User's change preserved
+    expect(config).toMatch(/sandbox_mode\s*=\s*['"]sandbox['"]/);
+    // Original key from base config still present
+    expect(config).toContain("preferred_auth_method");
+  });
+
+  test("idempotent-run-twice-no-change", async () => {
+    const { id, scripts } = await setup();
+    await runScripts(id, scripts);
+    const configAfterFirst = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+
+    // Second run without any user edits
+    await runScripts(id, scripts);
+    const configAfterSecond = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+
+    // Config should still contain the same keys
+    expect(configAfterSecond).toContain("preferred_auth_method");
+    // The values must match (quotes may change on roundtrip, compare via JSON)
+    const toJson = async (toml: string) => {
+      const resp = await execContainer(id, [
+        "bash",
+        "-c",
+        `echo '${toml.replace(/'/g, "'\\''")}' | dasel -i toml -o json`,
+      ]);
+      return JSON.parse(resp.stdout);
+    };
+    expect(await toJson(configAfterSecond)).toEqual(
+      await toJson(configAfterFirst),
+    );
+  });
+
+  test("idempotent-mcp-new-servers-added-existing-kept", async () => {
+    // First run: one MCP server
+    const mcpRun1 = [
+      "[mcp_servers.github]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-github"]',
+      'type = "stdio"',
+    ].join("\n");
+    const { id, scripts } = await setup({
+      moduleVariables: { mcp: mcpRun1 },
+    });
+    await runScripts(id, scripts);
+
+    // User adds their own MCP server manually
+    await execContainer(id, [
+      "bash",
+      "-c",
+      `cat >> /home/coder/.codex/config.toml << 'EOF'
+
+[mcp_servers.custom]
+command = "my-tool"
+args = ["--serve"]
+type = "stdio"
+EOF`,
+    ]);
+
+    // Second run: same module config
+    await runScripts(id, scripts);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // Module's github server still present
+    expect(config).toContain("github");
+    // User's manually-added custom server preserved
+    expect(config).toMatch(/command\s*=\s*['"]my-tool['"]/);
+  });
+
+  test("idempotent-ai-gateway-preserves-user-provider", async () => {
+    const { id, coderEnvVars, scripts } = await setup({
+      moduleVariables: {
+        enable_ai_gateway: "true",
+      },
+    });
+    await runScripts(id, scripts, coderEnvVars);
+
+    // User changes model_provider
+    await execContainer(id, [
+      "bash",
+      "-c",
+      "sed -i 's/model_provider = .*/model_provider = \"custom_provider\"/' /home/coder/.codex/config.toml",
+    ]);
+
+    // Second run
+    await runScripts(id, scripts, coderEnvVars);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // User's custom provider survives
+    expect(config).toMatch(/model_provider\s*=\s*['"]custom_provider['"]/);
+  });
+
+  test("base-config-plus-mcp-combined", async () => {
+    const baseConfig = [
+      'sandbox_mode = "danger-full-access"',
+      'preferred_auth_method = "apikey"',
+    ].join("\n");
+    const mcpConfig = [
+      "[mcp_servers.github]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-github"]',
+      'type = "stdio"',
+    ].join("\n");
+    const { id, scripts } = await setup({
+      moduleVariables: {
+        base_config_toml: baseConfig,
+        mcp: mcpConfig,
+      },
+    });
+    await runScripts(id, scripts);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // Base config keys present
+    expect(config).toContain("sandbox_mode");
+    expect(config).toContain("preferred_auth_method");
+    // MCP server present
+    expect(config).toContain("mcp_servers");
+    expect(config).toContain("github");
+  });
+
+  test("all-config-sources-combined", async () => {
+    const baseConfig = [
+      'sandbox_mode = "danger-full-access"',
+      'preferred_auth_method = "apikey"',
+    ].join("\n");
+    const mcpConfig = [
+      "[mcp_servers.github]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-github"]',
+      'type = "stdio"',
+    ].join("\n");
+    const { id, coderEnvVars, scripts } = await setup({
+      moduleVariables: {
+        enable_ai_gateway: "true",
+        base_config_toml: baseConfig,
+        mcp: mcpConfig,
+      },
+    });
+    await runScripts(id, scripts, coderEnvVars);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // Base config
+    expect(config).toContain("sandbox_mode");
+    expect(config).toContain("preferred_auth_method");
+    // MCP
+    expect(config).toContain("github");
+    // AI gateway
+    expect(config).toContain("model_providers");
+  });
+
+  test("idempotent-all-sources-user-edits-survive", async () => {
+    const baseConfig = [
+      'sandbox_mode = "danger-full-access"',
+      'preferred_auth_method = "apikey"',
+    ].join("\n");
+    const mcpConfig = [
+      "[mcp_servers.github]",
+      'command = "npx"',
+      'args = ["-y", "@modelcontextprotocol/server-github"]',
+      'type = "stdio"',
+    ].join("\n");
+    const { id, coderEnvVars, scripts } = await setup({
+      moduleVariables: {
+        enable_ai_gateway: "true",
+        base_config_toml: baseConfig,
+        mcp: mcpConfig,
+      },
+    });
+    await runScripts(id, scripts, coderEnvVars);
+
+    // User edits multiple things
+    await execContainer(id, [
+      "bash",
+      "-c",
+      [
+        "CONFIG=/home/coder/.codex/config.toml",
+        // Change auth method
+        "sed -i \"s/preferred_auth_method.*/preferred_auth_method = 'oauth'/\" $CONFIG",
+        // Add a custom top-level key
+        "echo 'user_note = \"do not touch\"' >> $CONFIG",
+      ].join(" && "),
+    ]);
+
+    // Second run
+    await runScripts(id, scripts, coderEnvVars);
+    const config = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+    // User edits survived
+    expect(config).toMatch(/preferred_auth_method\s*=\s*['"]oauth['"]/);
+    expect(config).toContain("user_note");
+    // Module config still present
+    expect(config).toContain("sandbox_mode");
+    expect(config).toContain("github");
+    expect(config).toContain("model_providers");
+  });
+
   test("custom-config-drops-reasoning-effort", async () => {
     const baseConfig = [
       'sandbox_mode = "danger-full-access"',
--- a/registry/coder-labs/modules/codex/main.test.ts
+++ b/registry/coder-labs/modules/codex/main.test.ts
@@ -527,26 +527,26 @@ EOF`,
   test("idempotent-run-twice-no-change", async () => {
     const { id, scripts } = await setup();
 
-    // Convert the container config file to JSON for comparison
-    const configToJson = async () => {
-      const resp = await execContainer(id, [
-        "bash",
-        "-c",
-        "dasel -i toml -o json < /home/coder/.codex/config.toml",
-      ]);
-      return JSON.parse(resp.stdout);
-    };
-
     // First run
     await runScripts(id, scripts);
-    const jsonAfterFirst = await configToJson();
 
-    // Second run without any user edits
+    // Second run triggers a dasel roundtrip (quotes may change)
     await runScripts(id, scripts);
-    const jsonAfterSecond = await configToJson();
+    const configAfterSecond = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
 
-    // Data must be identical (quotes may differ, so compare JSON not strings)
-    expect(jsonAfterSecond).toEqual(jsonAfterFirst);
+    // Third run: if idempotent, output must be identical to second run
+    await runScripts(id, scripts);
+    const configAfterThird = await readFileContainer(
+      id,
+      "/home/coder/.codex/config.toml",
+    );
+
+    // After the first roundtrip the serialization is stable, so a byte
+    // comparison is valid from the second run onward.
+    expect(configAfterThird).toEqual(configAfterSecond);
   });
 
   test("idempotent-mcp-new-servers-added-existing-kept", async () => {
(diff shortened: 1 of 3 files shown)


registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -111,58 +143,69 @@ function install_codex() {
...
   local config_path="$HOME/.codex/config.toml"
   mkdir -p "$(dirname "$${config_path}")"
 
+  # Build the new config entirely in JSON.
+  local new_json
+
   if [ -n "$${ARG_BASE_CONFIG_TOML}" ]; then
     printf "Using provided base configuration\n"
-    echo "$${ARG_BASE_CONFIG_TOML}" > "$${config_path}"
+    new_json=$(echo "$${ARG_BASE_CONFIG_TOML}" | dasel -i toml -o json)
```
matifali: why are we mixing json and toml?

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -111,58 +143,69 @@ function install_codex() {
...
   fi
 
   if [ "$${ARG_ENABLE_AI_GATEWAY}" = "true" ] && [ -n "$${ARG_AIBRIDGE_CONFIG}" ]; then
-    if ! grep -q '\[model_providers\.aigateway\]' "$${config_path}" 2>/dev/null; then
-      printf "Adding AI Gateway configuration\n"
-      echo -e "\n$${ARG_AIBRIDGE_CONFIG}" >> "$${config_path}"
-    else
-      printf "AI Gateway provider already defined in config, skipping append\n"
+    printf "Adding AI Gateway configuration\n"
+    local aibridge_json
```
matifali: we do not use aibridge anymore. rename to 
```suggestion
    local ai_gateway_json
```

35C4n0r: @matifali, is it okay if we release it in a new patch ? The readme already seems to be using ai_gateway for user facing variables, I'll do internal refactor in a new PR.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: `dasel` does not support deep merge of toml, so I came up with the idea that all the internal processing will be done in json (jq ftw), the final output will be written in toml. `dasel` only does 2 things 1. convert any toml to json for internal manipulation, 2. write the final json output as toml.

matifali (dismissed):
why are we doing toml to json to toml? a few real costs with this:

- comments in `~/.codex/config.toml` get stripped on every restart ([dasel](https://github.com/TomWright/dasel) can't preserve them).
- two new binary deps ([dasel](https://github.com/TomWright/dasel) + [jq](https://github.com/jqlang/jq)), and the dasel download isn't checksum-verified.

two cleaner options:

**marker block**: module owns a fenced region, user owns everything outside.

```toml
# >>> coder-managed: codex module >>>
preferred_auth_method = "apikey"
[mcp_servers.github]
command = "npx"
# <<< coder-managed: codex module <<<
```

`sed` strips and re-emits the block on each run. no deps, comments preserved, byte-stable. overrides go through terraform variables, which matches the rest of the registry.

**[`yq`](https://github.com/mikefarah/yq) (mikefarah) native toml**: if we keep merging, `yq -p toml -o toml eval-all '. as $i ireduce ({}; . *+ $i)' a.toml b.toml` is the same thing in one binary instead of two. still loses comments but a strict win over dasel + jq.

prefer marker block. happy to be wrong if there's a reason.

35C4n0r:
yq dosen't seem to work.

35C4n0r:
as for the block marker, everything written by the module will go inside the block marker (this includes the mcps provided by the user and the base_toml or user_provided_toml).
user/codex will append outside of it. 
also to be noted that if the user/codex modifies the section in the block marker, it will be overwritten on the next restart.

35C4n0r (commented):
The marker-block approach is a solid design: switching from destructive overwrite to idempotent merge is the right call, and the atomic write (temp-file then `mv`) is good practice. The test suite has good breadth.

Two correctness bugs in the script and two test gaps need attention before merge. 2 P1 (script bugs), 2 P1 (test gaps), 4 P2, 2 P3, 2 nits across 10 inline comments.

*Generated by Coder Agents (deep-review)*

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
     fi
   fi
+
+  # Preserve user content from the existing config.
+  local user_content=""
+  if [ -s "$${config_path}" ]; then
+    if grep -qF "$${MANAGED_START}" "$${config_path}"; then
+      # Strip the managed block; everything else is user content.
+      user_content=$(sed '/# >>> coder-managed: codex module >>>/,/# <<< coder-managed: codex module <<</d' "$${config_path}")
+    else
```
35C4n0r: **P1** `sed` range delete silently destroys all user content below the block when `MANAGED_END` is absent. *(Edge Case Analyst P1, Contract Auditor P1, Structural Analyst P1)*

> The strip command is `sed '/# >>> coder-managed.../,/# <<< coder-managed.../d'`. GNU sed applies a `/start/,/end/` range to EOF when the end pattern is never found. If a user hand-edits the config and accidentally removes or truncates the `MANAGED_END` line, the next workspace start silently deletes every user section appended below the managed block. `set -euo pipefail` gives no warning; user data is unrecoverable without a backup.

Fix: add a guard before this call. Check that `MANAGED_END` also exists in the file before running the sed strip; if only START is present, treat the file as truncated and skip stripping (or abort with an error). Something like `grep -qF "$MANAGED_END"` in the same `if` branch.

---

**P2** `sed` hardcodes the marker strings as literals while `grep` correctly uses `$MANAGED_START`/`$MANAGED_END`. *(Edge Case Analyst P2, Structural Analyst P2, Contract Auditor P3)*

> Line 580 uses `grep -qF "$MANAGED_START"` (reads the variable). Line 582 passes a hardcoded literal to `sed`. If the marker text is ever changed, `grep` will detect the new marker and conclude the file is managed, but `sed` will fail to match the old literal and return the full file unmodified as `user_content`, including the stale managed block. The script continues with no error and assembles a doubled managed block.

Fix: either use the variables in `sed` (with appropriate escaping, since sed addresses don't need the special chars here) or add a comment explaining that the duplication is intentional.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
+      # No markers (pre-upgrade or hand-written config). Treat the whole file as user content.
+      user_content=$(cat "$${config_path}")
+    fi
```
35C4n0r, line 199: **P1** Pre-upgrade bare-key hoisting produces duplicate TOML root-scope keys, making the assembled config invalid per spec. *(Structural Analyst P0, Edge Case Analyst P2, Contract Auditor P2)*

> When the existing file has no markers, all content becomes `user_content`, and bare keys (lines before the first `[`) are written to `user_bare`, placed above the managed block. The managed block then writes its own version of those same keys. For example, a legacy config with `preferred_auth_method = "login"` results in the key appearing twice at root scope: once from `user_bare` and once from `write_minimal_default_config`. TOML 1.0 §3 states duplicate keys are invalid; spec-compliant parsers reject the file entirely.

> *Test Auditor adds:* `idempotent-no-markers-preserves-user-config` seeds the file with `preferred_auth_method = "login"` and confirms it appears before `MANAGED_START`, but never asserts the key is absent from the managed block, so the test passes while shipping an invalid config.

Fix: when assembling `user_bare`, strip any key that is also emitted by the managed block, or document that user bare keys take full precedence and suppress the managed-block key for any key already present in `user_bare`.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
+
+  # Build managed-block content in a temp file.
+  local managed
+  managed=$(mktemp)
+
   if [ -n "$${ARG_BASE_CONFIG_TOML}" ]; then
     printf "Using provided base configuration\n"
-    echo "$${ARG_BASE_CONFIG_TOML}" > "$${config_path}"
+    printf '%s\n' "$${ARG_BASE_CONFIG_TOML}" > "$${managed}"
   else
```
35C4n0r, line 159: **P2** `ARG_BASE_CONFIG_TOML` and `ARG_MCP` inputs are not sanitized against the marker strings. *(Edge Case Analyst P2)*

> If either input contains `# >>> coder-managed: codex module >>>`, the managed block will embed a spurious `MANAGED_START` marker. On the next run, `sed` matches `START` at the outer header and `END` at the first `MANAGED_END` encountered, which may be the one embedded inside the base config rather than the actual block end, truncating the real managed content and treating the remainder as user content. This produces non-converging oscillation between runs.

Fix: add a check before writing `$ARG_BASE_CONFIG_TOML` or `$ARG_MCP` to `$managed`: fail with a clear error if either contains the marker string.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
+  fi
+
+  # Split user content into bare keys (before any [section]) and sections.
+  # Bare keys go above the managed block so they stay at TOML root scope;
+  # sections go below it where each [header] resets scope naturally.
+  local user_bare="" user_sections=""
+  if [ -n "$${user_content}" ]; then
+    local first_section
+    first_section=$(printf '%s\n' "$${user_content}" | grep -n '^\[' | head -1 | cut -d: -f1)
+    if [ -n "$${first_section}" ]; then
```
35C4n0r: **P2** Comments immediately before a `[section]` header are split from their section and placed on the opposite side of the managed block. *(Contract Auditor P2)*

> `first_section` marks the index of the first `^[` in `user_content`. Everything before it goes into `user_bare` (above the managed block); the section itself goes into `user_sections` (below). A comment such as `# My personal MCP server` on the line before `[mcp_servers.notes]` therefore ends up above `MANAGED_START`, disconnected from the section it describes. Test `idempotent-user-comments-preserved` acknowledges this in a code comment ("Bare-key comment hoisted above managed block") but only asserts `toContain(...)`, so it passes without detecting the visual disconnection.

Fix: scan for the first `[` line that is not preceded on the same line by content, and treat the lines immediately before it (comments + blank lines) as the start of the section rather than bare scope.

---

**P3** `grep -n '^\['` is not a reliable TOML section header detector. *(Structural Analyst P3)*

> Any non-header line that begins with `[` (e.g., an array value written as `key = [\n]` on separate lines, or a multi-line basic string continuation) causes the bare/section split to happen at the wrong boundary, silently misclassifying post-section content as root-scope bare keys.

This is lower risk in practice (Codex configs are simple), but worth noting for the future.

---

**Nit** `first_section` holds a line number (an integer), not the section's content. *(Style Reviewer)*

`first_section_lineno` or `first_section_line` would match what the variable actually holds and is immediately used as: `head -n "$((first_section - 1))"`.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
+  {
+    if [ -n "$${user_bare}" ]; then
+      printf '%s\n\n' "$${user_bare}"
+    fi
+    printf '%s\n' "$${MANAGED_START}"
+    cat "$${managed}"
+    printf '%s\n' "$${MANAGED_END}"
+    if [ -n "$${user_sections}" ]; then
+      printf '\n%s\n' "$${user_sections}"
+    fi
```
35C4n0r, line 216: **P2** `user_sections` are appended without checking for table name collisions with the managed block. *(Structural Analyst P2)*

> `user_sections` are written after `MANAGED_END` with no collision check. If a user appends `[model_providers.aigateway]` below the markers, and `enable_ai_gateway = true` also puts `[model_providers.aigateway]` in the managed block, both table definitions land in the final file. The TOML spec forbids redefining a table; this produces a structurally invalid config, silently ignored by the script.

Fix: either document clearly in the README that user sections must not reuse managed table names, or add a validation pass that detects duplicate table headers and warns or errors.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
+  fi
+
+  # Assemble final config atomically:
+  #   user bare keys  (root scope)
+  #   managed block   (bare keys first, then [sections])
+  #   user sections   (each [header] resets scope)
+  local tmp
+  tmp=$(mktemp "$${config_path}.XXXXXX")
+  {
```
35C4n0r, line 207: **P3** No `trap` for temp file cleanup; both `$managed` and `$tmp` are leaked when `set -e` fires before the function returns. *(Edge Case Analyst P3, Contract Auditor P3)*

> `managed=$(mktemp)` creates a file in `/tmp`; `tmp=$(mktemp "$config_path.XXXXXX")` creates a sibling of `config.toml` in `~/.codex/`. The only cleanup is `rm -f "$managed"` at the end of the function, which is skipped if any earlier command exits non-zero. On repeated provisioning failures (disk-full, permission denied), `config.toml.XXXXXX` files accumulate in `~/.codex/`.

Fix: add `trap 'rm -f "${managed:-}" "${tmp:-}"' EXIT` immediately after both `mktemp` calls.

---

**Nit** `managed` and `tmp` are both `mktemp` results in the same function; `tmp` is named by type while `managed` is named by role. *(Style Reviewer)*

`config_tmp` would make the atomic-swap purpose explicit and make `rm -f "$managed"` clearly not the last cleanup needed.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -114,27 +114,84 @@ function populate_config_toml() {
...
     else
       printf "AI Gateway provider already defined in config, skipping append\n"
     fi
   fi
+
+  # Preserve user content from the existing config.
+  local user_content=""
+  if [ -s "$${config_path}" ]; then
+    if grep -qF "$${MANAGED_START}" "$${config_path}"; then
+      # Strip the managed block; everything else is user content.
```
35C4n0r: **P3** The config file is read twice (`grep` then `sed`) without any locking, creating a TOCTOU window. *(Security Reviewer P3)*

> Line 42 (`grep -qF`) snapshots whether the marker is present; line 44 (`sed`) reads the file again to extract user content. A concurrent workspace restart or a user-side script modifying the file between these two reads can cause `sed` to operate on a different file state than `grep` observed. The atomic `mv` at the end prevents a torn write, but content written after the `sed` snapshot is silently discarded. Coder workspaces can restart quickly or in parallel, making this a realistic race.

The fix requires either holding a file descriptor open across both reads or using a lock file. Holding the `sed` output from a single `< "$config_path"` read would be the simplest approach.

registry/coder-labs/modules/codex/main.test.ts
```diff
@@ -441,7 +505,341 @@ describe("codex", async () => {
...
+  test("idempotent-managed-block-regenerated", async () => {
+    const { id, scripts } = await setup({
+      moduleVariables: {
+        model_reasoning_effort: "high",
+      },
+    });
+    await runScripts(id, scripts);
+
+    // User modifies a value inside the managed block.
+    await execContainer(id, [
```
35C4n0r, line 598: **P1** `idempotent-user-section-survives-restart` checks that `[mcp_servers.user_tool]` exists anywhere in the file but never asserts it falls *after* `MANAGED_END`. *(Test Auditor P1)*

> If a bug placed the user section inside or before the managed block, all assertions still pass. The sibling test `idempotent-user-bare-keys-stay-at-root-scope` (added in this same PR) shows the correct pattern: take `config.indexOf(MANAGED_END)` and `config.indexOf("[mcp_servers.user_tool]")`, then `expect(sectionIdx).toBeGreaterThan(endIdx)`. This is the primary invariant the test claims to enforce.

Add: `const endIdx = config.indexOf(MANAGED_END); const sectionIdx = config.indexOf("[mcp_servers.user_tool]"); expect(sectionIdx).toBeGreaterThan(endIdx);`

registry/coder-labs/modules/codex/main.test.ts
```diff
@@ -441,7 +505,341 @@ describe("codex", async () => {
...
+    // Module's github server still present (in managed block)
+    expect(config).toContain("[mcp_servers.github]");
+    expect(config).toMatch(/command\s*=\s*"npx"/);
+    // User's custom server preserved (outside managed block)
+    expect(config).toContain("[mcp_servers.custom]");
+    expect(config).toMatch(/command\s*=\s*"my-tool"/);
+  });
+
+  test("idempotent-no-markers-preserves-user-config", async () => {
+    const { id, scripts } = await setup();
```
35C4n0r, line 737: **P1** `idempotent-no-markers-preserves-user-config` simulates a legacy file and runs the script once, but never does a second run. *(Test Auditor P1)*

> The migration from legacy-to-marker format is only half-tested: the test does not verify that the migrated file is itself idempotent on a subsequent restart. A bug where run 2 double-hoists the already-hoisted bare keys, or produces duplicate markers, is invisible. Add a second `runScripts` call and assert the output is byte-identical to the first.

This is especially important given the P1 duplicate-key bug on this path.

registry/coder-labs/modules/codex/main.test.ts
```diff
@@ -441,7 +505,341 @@ describe("codex", async () => {
...
+    // Bare-key comment hoisted above managed block
+    expect(config).toContain("# My personal top-level setting");
+    // Section comments preserved below managed block
+    expect(config).toContain("# My personal MCP server");
+    expect(config).toContain("# This server is for my personal notes");
+    expect(config).toContain("[mcp_servers.notes]");
+  });
+
+  test("idempotent-stable-after-roundtrip", async () => {
+    const { id, scripts } = await setup();
```
35C4n0r, line 658: **P2** `idempotent-stable-after-roundtrip` would pass even if the marker-block feature were deleted entirely. *(Test Auditor P2)*

> The test runs twice on a clean container with no user content added between runs. A naive "always overwrite the file" implementation also produces byte-identical output in this scenario because the inputs are deterministic. The test proves only that the script is deterministic, which is a much weaker property than idempotency of the marker-block feature. Add user content between the two runs (as `idempotent-multiple-restarts-user-content-stable` does) to make the assertion meaningful.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Fixed in [220ffa9](https://github.com/coder/registry/commit/220ffa9). Added a `grep -qF "$MANAGED_END"` check inside the `MANAGED_START`-found branch. If the end marker is absent the function now logs a warning and treats the entire file as user content rather than running the range delete, so no user data is lost.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Addressed in [3007099](https://github.com/coder/registry/commit/3007099) by changing strategy: when no markers are found (first run / pre-upgrade), `user_content` stays empty and the file is overwritten cleanly by the managed block. There is no bare-key hoisting, so the duplicate-root-key scenario cannot occur. The deduplication block added in round 1 was removed as it's no longer needed.

registry/coder-labs/modules/codex/main.test.ts
35C4n0r: Fixed in [220ffa9](https://github.com/coder/registry/commit/220ffa9). The test now takes `endIdx = config.indexOf(MANAGED_END)` and `sectionIdx = config.indexOf("[mcp_servers.user_tool]")` and asserts `expect(sectionIdx).toBeGreaterThan(endIdx)`.

registry/coder-labs/modules/codex/main.test.ts
35C4n0r: Fixed across [220ffa9](https://github.com/coder/registry/commit/220ffa9) and [3007099](https://github.com/coder/registry/commit/3007099). The test is renamed `no-markers-first-run-overwrites` and now runs the script twice: the second `runScripts` call is followed by `expect(configAfterSecond).toEqual(config)` to confirm the output is byte-stable. The assertion set also changed: since the first run now does a full overwrite (no hoisting), the test asserts legacy content is *absent* after run 1 rather than preserved above the managed block.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Fixed in [c2360d0](https://github.com/coder/registry/commit/c2360d0). Added a `for _arg in "${ARG_BASE_CONFIG_TOML}" "${ARG_MCP}"` loop right after the marker declarations that pipes each non-empty input through `grep -qF "${MANAGED_START}"` and returns 1 with an error message if the marker is found.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Fixed in [c2360d0](https://github.com/coder/registry/commit/c2360d0).

**P2 (comment splitting):** Added a backward-walk `while` loop that decrements `first_section_lineno` past consecutive comment (`#`) and blank lines before committing the split point. Comments immediately preceding a `[section]` header now stay in `user_sections` rather than being stranded in `user_bare`.

**P3 (`grep '^\['` reliability):** Added an inline comment noting this may match a bare `[` at the start of a multi-line array literal, and that it is low risk for Codex configs which don't use that syntax.

**Nit:** `first_section` renamed to `first_section_lineno`.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Addressed in [c2360d0](https://github.com/coder/registry/commit/c2360d0) with a README note rather than a logic change. Added a `> [!NOTE]` block in the Configuration section warning that `[section]` headers placed below `MANAGED_END` must not reuse a table name already defined inside the managed block, and that duplicate table definitions will cause Codex to reject the config.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Fixed in [c2360d0](https://github.com/coder/registry/commit/c2360d0). Both temp variables are now declared as `local managed="" config_tmp=""` upfront, and `trap 'rm -f "${managed:-}" "${config_tmp:-}"' EXIT` is set immediately after — so any early exit from `set -e` cleans both files. `tmp` is renamed to `config_tmp` throughout to make the atomic-swap purpose explicit.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: Fixed in [c2360d0](https://github.com/coder/registry/commit/c2360d0). The file is now read once with `existing_config=$(cat "${config_path}")` behind the `[ -s ]` guard. All subsequent marker checks and the strip operation pipe from `printf '%s\n' "${existing_config}"` rather than re-opening the file, collapsing the three separate reads into one.

registry/coder-labs/modules/codex/main.test.ts
35C4n0r: Fixed in [c2360d0](https://github.com/coder/registry/commit/c2360d0). The test now adds a bare key and a commented section after run 1, then does a second run (captures the config with user content in place) and a third run, asserting `configAfterThird.toEqual(configAfterSecond)` plus that both user items are still present. A naive always-overwrite implementation would fail here because the user content would be gone after run 2.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
35C4n0r: we use marker strings now.

35C4n0r:
@andrewdennis117 PR is ready for review, I'll do a version bump in README once the code review passes.

registry/coder-labs/modules/codex/scripts/install.sh.tftpl
```diff
@@ -136,27 +136,86 @@ function populate_config_toml() {
...
+    if [ -n "$${user_before}" ]; then
+      printf '%s\n\n' "$${user_before}"
+    fi
+    printf '%s\n' "$${MANAGED_START}"
+    cat "$${managed}"
+    printf '%s\n' "$${MANAGED_END}"
+    if [ -n "$${user_after}" ]; then
+      printf '\n%s\n' "$${user_after}"
+    fi
+  } > "$${config_tmp}" && mv "$${config_tmp}" "$${config_path}"
```
andrewdennis117, line 217: Would you consider adding a `grep`-based warning here (stderr, non-fatal) that flags when a `[section]` header in `user_after` matches a header inside the managed block? Cheap to add and saves a debugging session.

It's fine if you do not think this is helpful. I am good to approve without.

35C4n0r: @andrewdennis117 having that as a feature would have been helpful, and it came in one of the deep reviews too. The reason I haven't added is:
- It will increase the complexity of the script, we could have done this if we had a good toml parser on hand.
- We have added this note in the readme.md, so we should be good to go:
>`[section]` headers you place **below** the managed block must not duplicate a table name already defined inside it (same reason). Bare keys placed after the managed block will fall under whatever `[section]` preceded them — place bare keys above the block to keep them at root scope.

andrewdennis117 (approved):

PR #904, opened 2026-05-28
fix(coder/modules/dotfiles): allow tilde in DOTFILES_URI shell validation

### LINKED ISSUE #762: dotfiles: tilde sign in git repository URL is not allowed
### Problem
Since dotfiles v1.2.4, the git repository URL is validated using a regex. In my company's Bitbucket setup, the user repository URLs always include a tilde (~) sign, e. g. like that:
`ssh://git@bitbucket.mycompany.org:7999/~myusername/example-repository.git`

The tilde is not allowed in the regex, rendering the dotfiles plugin v1.2.4 and newer useless for us.

### Proposed solution
The validation regex should be changed to allow using "~".

35C4n0r:
Fixes https://github.com/coder/registry/issues/762

## Problem

The shell-side URI validation regex in `run.sh` did not include `~` in the allowed character set. URLs containing tilde paths (common in Bitbucket Server personal repositories, e.g. `ssh://git@bitbucket.example.org:7999/~user/repo.git`) were rejected at runtime with `ERROR: DOTFILES_URI contains invalid characters`.

The Terraform-side validations in `main.tf` already allowed `~`, so the inconsistency only surfaced when the script actually ran in a workspace.

## Changes

- **`run.sh`**: Added `~` to the character class in the shell validation regex, making it consistent with the three Terraform regex patterns in `main.tf`.
- **`main.test.ts`**: The "accepts valid git URL formats" test now also executes the rendered shell script and asserts that the shell-side validation does not reject any of the valid URLs. This closes the coverage gap that let the Terraform/shell inconsistency go undetected.

> 🤖 Generated by Coder Agents

### DIFF
--- a/registry/coder/modules/dotfiles/README.md
+++ b/registry/coder/modules/dotfiles/README.md
@@ -18,7 +18,7 @@ Under the hood, this module uses the [coder dotfiles](https://coder.com/docs/v2/
 module "dotfiles" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/dotfiles/coder"
-  version  = "1.4.1"
+  version  = "1.4.2"
   agent_id = coder_agent.example.id
 }
 ```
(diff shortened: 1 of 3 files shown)


matifali (approved):
LGTM.

@35C4n0r please test manually before merging.

PR #907, opened 2026-06-02
fix(coder-labs/gemini): base64-encode task prompt to handle special characters

jdomeracki-coder:
Encodes the Gemini module's `task_prompt` before passing it into the start script and decodes it in `start.sh`, matching how `gemini_system_prompt` is already handled. Prompts containing single quotes or other shell metacharacters are now passed through correctly instead of breaking the start script.

## Type of Change

- [x] Bug fix

## Module Information

**Path:** `registry/coder-labs/modules/gemini`  
**New version:** `v3.0.2`  
**Breaking change:** No

## Testing & Validation

- [x] Tests pass (`bun test main.test.ts`) — added a regression test for prompts containing special characters
- [x] Changes tested locally

## Related Issues

None

---
<sub>Generated by Coder Agents on behalf of @jdomeracki-coder.</sub>

### DIFF
--- a/registry/coder-labs/modules/gemini/README.md
+++ b/registry/coder-labs/modules/gemini/README.md
@@ -13,7 +13,7 @@ Run [Gemini CLI](https://github.com/google-gemini/gemini-cli) in your workspace 
 ```tf
 module "gemini" {
   source   = "registry.coder.com/coder-labs/gemini/coder"
-  version  = "3.0.1"
+  version  = "3.0.2"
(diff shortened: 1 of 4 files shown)


DevelopmentCats (approved):
LGTM 

I will test locally just to ensure it behaves the same way but I cant see why it wouldn't

DevelopmentCats:
Behaves fine! Merging and Releasing.

PR #923, opened 2026-06-13
fix(vscode-web): point app healthcheck at /version instead of /healthz

### LINKED ISSUE #922: modules/vscode-web: healthcheck probes /healthz, which the VS Code web server does not expose
## Problem

The `coder_app` healthcheck of the vscode-web module points at `/healthz`:

```hcl
healthcheck_url = var.subdomain ? "http://localhost:${var.port}/healthz" : "http://localhost:${var.port}${local.server_base_path}/healthz"
```

Microsoft's standalone VS Code web server (the `vscode-server-linux-*-web` build this module downloads and runs via `serve-local`) has no `/healthz` route. Against the current stable build (1.124.2):

```
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:13338/healthz
404
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:13338/version
200
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:13338/
200
```

`/healthz` presumably carried over from the code-server module — code-server really does expose `/healthz`, but the Microsoft server does not.

## Impact

Mostly correctness/clarity rather than a visible breakage: the agent treats any response below 500 as healthy ([`agent/apphealth.go`](https://github.com/coder/coder/blob/main/agent/apphealth.go)), so the 404 never shows up as an unhealthy app. But the check is effectively only "does anything answer HTTP on that port", and the URL misleads anyone copying the module's pattern into their own templates or `customizations.coder.apps` (where I ran into it: coder/devcontainer-features#22).

## Suggested fix

Probe `/version` instead, which the web server answers with 200 in both the subdomain and path-based cases. I'll send a PR.

schnell3526:
Closes #922.

## Summary

The vscode-web module's `coder_app` healthcheck probes `/healthz`, but Microsoft's standalone VS Code web server exposes no such route — it returns 404 (the path was presumably carried over from the code-server module, where code-server really does serve `/healthz`). Against the current stable build (1.124.2):

```
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:13338/healthz
404
$ curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:13338/version
200
```

The 404 never surfaced in dashboards because the agent treats any response below 500 as healthy (`agent/apphealth.go`), but as-is the check only verifies that something answers HTTP on the port, and the URL misleads anyone copying the pattern into their own templates.

This switches the healthcheck to `/version`, which returns 200 in both the subdomain and path-based branches.

## Changes

- `main.tf`: `healthcheck_url` probes `/version` instead of `/healthz`
- `README.md`: patch version bump via `./.github/scripts/version-bump.sh patch`

Found while building a vscode-web dev container feature (coder/devcontainer-features#22).

### DIFF
--- a/registry/coder/modules/vscode-web/README.md
+++ b/registry/coder/modules/vscode-web/README.md
@@ -14,7 +14,7 @@ Automatically install [Visual Studio Code Server](https://code.visualstudio.com/
 module "vscode-web" {
   count          = data.coder_workspace.me.start_count
   source         = "registry.coder.com/coder/vscode-web/coder"
-  version        = "1.5.0"
+  version        = "1.5.1"
   agent_id       = coder_agent.example.id
   accept_license = true
 }
@@ -30,7 +30,7 @@ module "vscode-web" {
 module "vscode-web" {
   count          = data.coder_workspace.me.start_count
   source         = "registry.coder.com/coder/vscode-web/coder"
-  version        = "1.5.0"
+  version        = "1.5.1"
   agent_id       = coder_agent.example.id
   install_prefix = "/home/coder/.vscode-web"
   folder         = "/home/coder"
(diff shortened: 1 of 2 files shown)


matifali (approved):
Thanks for your contribution

PR #934, opened 2026-06-26
chore(zed): make coder_script conditional on settings input

matifali:
The `coder_script` resource for configuring Zed settings was always created, even when no settings were provided. The script handled this at runtime with an early exit check, but it's cleaner to not create the resource at all.

## Changes

- Add `count = var.settings != "" ? 1 : 0` to `coder_script "zed_settings"` so it's only created when settings are provided
- Remove the redundant empty `SETTINGS_B64` runtime check from the shell script
- Update `zed.tftest.hcl` to use indexed reference (`coder_script.zed_settings[0]`) and assert no resource is created when settings is empty
- Update `main.test.ts` to verify no `coder_script` exists in state when settings is empty

This follows the same pattern used in cursor, windsurf, kiro, and antigravity modules.

> 🤖 This PR was created with the help of Coder Agents, and needs a human review.  🧑💻

### DIFF
--- a/registry/coder/modules/zed/README.md
+++ b/registry/coder/modules/zed/README.md
@@ -19,7 +19,7 @@ Zed is a high-performance, multiplayer code editor from the creators of Atom and
 module "zed" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/zed/coder"
-  version  = "1.1.4"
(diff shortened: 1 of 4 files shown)


35C4n0r (approved):
LGTM!

PR #1012, opened 2026-07-14
feat(.github): add recurring module scorecard workflow

bpmct:
Adds a recurring workflow that scores every module in the `coder` namespace against a documented rubric and posts the results as GitHub Discussions: one dedicated discussion per module, plus a pinned [📊 Module Scorecards](https://github.com/coder/registry/discussions/1011) index grouped by module type (Agent / IDE / Utility). A separate PR workflow compares changed modules against their current score and comments with the diff.

**Why a scorecard?** It gives us a shared definition of quality (and a goalpost) for modules: what we care about for each one, in one place, for any contributor. It is not a strict gate today, though it may become one, and the criteria will evolve over time. It also helps catch regressions, for example a feature or documented capability being removed, since scores are re-evaluated whenever a module changes.

The rubric lives in `.github/scorecard/SCORECARD.md` and is read directly by the scoring script at runtime, so editing that one file changes the criteria for every future run. No script changes needed.

## Suggested review order

1. `.github/scorecard/SCORECARD.md` - the contract (80 lines)
2. `.github/scorecard/lib.ts` - shared parsing + GraphQL helpers
3. `.github/scorecard/score-modules.ts` - scoring + discussion upserts + PR reports
4. `.github/scorecard/scorecard-index.ts` - index aggregation
5. The two workflow files - plumbing

Rather than simulating the prompt, validate against live output: all 48 modules were scored and posted from this code. Example module discussion: [Claude Code module](https://github.com/coder/registry/discussions/970); index: [#1011](https://github.com/coder/registry/discussions/1011).

## Changes

- `.github/scorecard/SCORECARD.md` - the rubric: universal themes (Presentation & Onboarding, Credential Hygiene, Restricted-Network Readiness, Engineering Quality) plus a track section (Agent / IDE / Utility) out of 100
- `.github/scorecard/lib.ts` - shared constants, summary-table parsing, GraphQL client, discussion lookup
- `.github/scorecard/score-modules.ts` - scores modules with Claude (`claude-sonnet-4-5`, temperature 0), recomputes the overall deterministically from theme scores, and upserts each module's discussion. `--pr-report` mode compares against the current discussion instead of posting
- `.github/scorecard/scorecard-index.ts` - rebuilds the pinned index from the module discussions, grouped by track
- `.github/workflows/module-scorecards.yaml` - weekly cron, `workflow_dispatch`, push to `main`: rubric change triggers a full run, module changes trigger a scoped run of just those modules; index rebuilds either way
- `.github/workflows/module-scorecard-check.yaml` - on PRs touching modules (internal only), scores the changed modules and posts a sticky comment flagging regressions, improvements, or unchanged scores. Read-only against discussions

## Secrets

Both are **already added** to this repo's Actions secrets, so the workflows are ready to run on merge:

- `SCORECARD_ANTHROPIC_API_KEY` - Anthropic API key for scoring
- `SCORECARD_DISCUSSIONS_TOKEN` - GitHub PAT with Discussions read/write

<details>
<summary>Design notes</summary>

- **Scoring**: each criterion is 0 / half / full. Full = implemented and documented. Criteria marked *if applicable* (Restricted-Network Readiness) are excluded from the denominator when the concern doesn't exist by construction (e.g. `git-clone` downloads nothing), and the score normalizes to 100.
- **Utility modules** skip the track section and normalize from the universal themes only.
- **Internal building blocks** (READMEs cautioning against direct use, e.g. `coder-utils`, `agentapi`, `vscode-desktop-core`) are excluded from scoring entirely via a deterministic README check.
- **Deterministic math**: the model emits per-criterion scores, but the overall is recomputed by the script from the summary row to prevent arithmetic drift. `temperature: 0` keeps re-scores stable (verified: identical re-score on an unchanged module).
- **Idempotency**: module discussions are found by title (display name + "module") and updated in place via a hidden `module-scorecard` HTML comment marker; the index and PR comments use `module-scorecard-index` and `module-scorecard-pr` markers respectively.
- **Display name renames**: the PR check falls back to finding the old discussion by registry URL and warns in its comment so a maintainer can retitle or delete the old one.
- **Index aggregation** reads the summary tables back from the module discussions, so scoped single-module runs still produce a complete index.
- **New modules** are picked up automatically: the PR check reports a "first scorecard", and the post-merge run creates the module's discussion and adds it to the index.

</details>

🤖 Generated with [Coder Agents](https://coder.com/docs/ai-coder/agents)


DevelopmentCats (approved):
I cant say that I have any complaints since we went through most of this together @bpmct  but Jakub might :P

david-fraley (commented):
Two primary thoughts for ya @bpmct 

1. Adding in a complementary skill for how to interact with all this, that points at `scorecard.md`. Could help with the inevitable AI use in module creation/maintenance that might occur. I'm thinking of this like school -- you, the prof, just handed out the rubric; can you give students a study sheet (the skill) so that they build what you want
2. Is it possible to extend this to templates?

PR #1029, opened 2026-07-27
ci: move setup-tf to official Hashicorp action

matifali:
The curent coder/coder action fails with

```console
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action 'coder/coder/.github/actions/setup-tf@5c2838a7fb93874eed2d4ab8e6b6676ae4a237ce'.
```

The upstream repo has removed this action and now uses mise in CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,8 +90,8 @@ jobs:
         with:
           bun-version: latest
       # Need Terraform for its formatter
-      - name: Install Terraform
-        uses: coder/coder/.github/actions/setup-tf@c3dd41e1e8c9d0a0860a9a81c3c050be55879dc2 # v2.34.5
+      - name: Set up Terraform
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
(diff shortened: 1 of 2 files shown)


35C4n0r:
@matifali version-bump.yaml too.

image

.github/workflows/ci.yaml
```diff
@@ -39,7 +39,9 @@ jobs:
             all:
               - '**'
       - name: Set up Terraform
-        uses: coder/coder/.github/actions/setup-tf@c3dd41e1e8c9d0a0860a9a81c3c050be55879dc2 # v2.34.5
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
+        with:
+          terraform_wrapper: false
```
35C4n0r, line 44: @matifali, can I know why the default true was failing ?

35C4n0r (approved):

.github/workflows/ci.yaml
matifali: Just shared the chat with you. It was false for the same reason in the previously used action too.

PR #1034, opened 2026-07-29
fix(windows-rdp): escape credentials and fix auto-connect on Gateway 2026.2.x

### LINKED ISSUE #20: bug: Windows-RDP - password field in Devolution Form is not populated correctly
Extension: Windows-RDP
Platform: Coder/Workspaces
Guest OS: Windows VM

**Description:**
When submitting passwords through the Devolution Form in the Windows-RDP extension, any backslash (\\) characters are omitted, resulting in incorrect password transmission.

**Steps to Reproduce:**

Include the Windows-RDP extension in a Workspace. Start it and use the 'Windows-RDP' Agent.

Observe that the password provided in the form is 'N;JVO*UmL^a*P', not 'N;JVO*U\mL^a*P' missing the backslash (using the :eye: to reveal the password) and compare it to the Workspace UI.

Submit the Devolutions form/login with the filled-in data (remote host is localhost, but shows as undefined, this does not make any difference to the exercise)

**Expected Behavior:**
The password should be transmitted identical from the Workspace UI password/user password, preserving all characters, including backslashes.

**Actual Behavior:**
The backslash character is omitted from the password upon submission, leading to authentication failures, InvalidToken: CredSSP ... STATUS_LOGON_FAILURE [0xc000006d]

**Impact:**
This issue prevents users from authenticating when their passwords contain backslashes, which are common in complex passwords.

**Potentially Affected Characters:**
While the backslash is confirmed to be affected, other special characters may also be mishandled. Characters that often require escaping and could be impacted are listed here as a need to test:
This will also be 'random issue' for users where their requirement of special characters and/or randomness of the generated password will sometimes include all working characters.

- Double quotes (")
- Single quotes (')
- Ampersands (&) and other 'HTML agnostic characters' such as % : \ $ @
- Less than (<) and greater than (>) symbols
- Pipes and redirection characters (| < >)
- Caret (^) OK
- Tilde (~)
- Programmable characters that may be affected in parsing data {} [], although curly braces were found in the example to work ok

**Suggested Investigation:**
Review the input handling and encoding mechanisms within the Devolution Form, particularly focusing on how special characters are processed and transmitted. Ensure that characters like backslashes are properly escaped or encoded to prevent omission or misinterpretation.

**Additional Notes:**
This issue aligns with generic problems where special characters in passwords cause authentication failures due to improper handling. For instance, a similar issue was reported where the ` (backtick) character in passwords led to login problems in RDP sessions .

This shows another potential issue where we might not always control the password or if the user does not use the random_password (which could exclude \ character) and a more sturdy method is needed.

### LINKED ISSUE #1044: bug: windows-rdp auto-fill no longer submits the form on Devolutions Gateway 2026.2.x
### Current Behavior

The `windows-rdp` module defaults to `devolutions_gateway_version = "latest"`, which currently installs Devolutions Gateway **2026.2.4**. The module's `devolutions-patch.js` selectors no longer match that UI, so the sign-in form is filled but never submitted, and no session starts.

Observed on a live AWS Windows workspace, browser console:

```
[Devolutions Patch] Form detected. Starting auto-fill...
[Devolutions Patch] Protocol dropdown trigger not found.
[Devolutions Patch] Hostname set to localhost
[Devolutions Patch] Set username to Administrator
[Devolutions Patch] Set password to <redacted>
[Devolutions Patch] Submit button not found or disabled.
ERROR TypeError: C.protocol.toLowerCase is not a function
    at A.socketOnMessage (.../chunk-AKNLUDAU.js)
```

The form stays on screen with the fields populated, the `session-toolbar` never appears, so `attachCloseListener`, `Unicode Keyboard Mode` and `Dynamic Resize` never get applied either.

Two selectors have drifted:

* `p-dropdown[id="protocol"]` no longer matches the protocol control
* `p-button[class="p-element"] button` no longer matches the connect button

### Expected Behavior

Opening the **Web RDP** app connects straight into the Windows desktop, with no visible form and no manual click.

### Steps to Reproduce

1. Add the `windows-rdp` module to a Windows workspace, leaving `devolutions_gateway_version` at its `latest` default.
2. Start the workspace and open the **Web RDP** app.
3. The Devolutions form appears pre-filled, but no session connects.

### Desired Solution

Update the selectors in `devolutions-patch.js` for the current Gateway UI, and make the failure loud rather than silent: if the submit button cannot be found after N polls, surface a visible message instead of only logging to the console. A user today just sees a form that does nothing.

Worth discussing separately: `latest` means an upstream UI change can break every workspace using this module with no change on our side. Pinning a known-good default and bumping it deliberately would trade freshness for predictability, which seems like the right trade for an air-gapped or otherwise change-averse deployment.

> [!NOTE]
> Found while dogfooding coder/registry#1034. Independent of the credential escaping fixed there.

Related to coder/registry#20

matifali:
## Problem

Two separate defects made the Web RDP session fail, and they compound: the credentials were mangled in transit, and even correct credentials were never submitted.

**1. Credentials were interpolated without escaping** (coder/registry#20)

* `devolutions-patch.js` embeds the password inside a double-quoted JS string literal (`value: "${CODER_PASSWORD}"`). JS parses the injected `\m` as an escape sequence and drops the backslash, so `N;JVO*U\mL^a*P` becomes `N;JVO*UmL^a*P`. A `"` breaks the patch script entirely.
* `powershell-installation-script.tftpl` passes the password as a PowerShell double-quoted string (`Set-AdminPassword -adminPassword "${admin_password}"`), where `$`, backtick, and `"` are interpreted. The Windows account then gets a different password than Coder displays.

Both paths end in `STATUS_LOGON_FAILURE [0xc000006d]`, and whether a user hits it depends on which characters their generated password happens to contain.

**2. The auto-fill selectors no longer match Gateway 2026.2.x** (coder/registry#1044)

The module defaults to `devolutions_gateway_version = "latest"`. On 2026.2.4 the form fills but never submits, and the in-session toolbar is never found, so the close-window listener never attaches and the poll loops forever.

## Changes

**Credential escaping**

| Destination | Encoding | Handles |
| -- | -- | -- |
| JS patch file | `jsonencode`, outer quotes trimmed | `\`, `"`, control chars, and `<` `>` `&` as `\u003c` style escapes |
| PowerShell script | single-quoted string, `'` doubled | `$`, backtick, `"`, `\`, and everything else literally |

The JS file keeps a real string literal, so it stays valid JavaScript for Prettier and `@ts-check`.

**Selector updates**

| Was | Now |
| -- | -- |
| `p-dropdown[id="protocol"]` | dropped, RDP is already the default and driving a PrimeNG overlay is what broke |
| `p-button[class="p-element"] button` | `button[type="submit"]` inside the form |
| `session-toolbar` | `floating-session-toolbar`, close button matched by `aria-label` |
| `setCheckbox("Unicode Keyboard Mode")` | removed, the option no longer exists |
| `setCheckbox("Dynamic Resize")` | `input#enableDisplayControl` on the connection form, corrected only when not already enabled |

Both polls are now bounded. On failure the form is revealed with a visible banner instead of looping forever and logging to a console the user never opens. Credential values are no longer logged to the browser console.

Module bumped to `1.3.1`.

## Validation

Provisioned a real AWS Windows workspace from a copy of `aws-windows-desktop`, with the module sourced from this branch and `admin_password = "Te\st$Pa\"ss'w`ord&<>|1!"\`.

Rendered output:

```
value: "Te\\st$Pa\"ss'w`ord\u0026\u003c\u003e|1!",
Set-AdminPassword -adminPassword 'Te\st$Pa"ss''w`ord&<>|1!'
```

On the VM, the account password round-trips exactly:

```powershell
$ctx.ValidateCredentials('Administrator', 'Te\st$Pa"ss''w`ord&<>|1!')   # True
$ctx.ValidateCredentials('Administrator', 'coderRDP!')                  # False
```

In the browser, unattended, no clicks:

```
[Devolutions Patch] Form detected. Starting auto-fill...
[Devolutions Patch] Hostname set to localhost
[Devolutions Patch] Set username
[Devolutions Patch] Set password
[Devolutions Patch] Form submitted.
[Devolutions Patch] Top bar detected. Proceeding with next steps...
[Devolutions Patch] Close listener attached.
canvas: 1   floating-session-toolbar: 1   failure banner: 0
```

`ironrdp web::session: Connected!`, live Windows desktop, no `STATUS_LOGON_FAILURE`.

`bun test main.test.ts` passes (5/5); `prettier` and `terraform fmt` are clean.

> [!NOTE]
> The same escaping bug exists in `amazon-dcv-windows` (coder/registry#1035) and `local-windows-rdp` (coder/registry#1043). Both are separate modules with separate releases, so they are filed rather than folded in here.

> [!CAUTION]
> `devolutions_gateway_version` still defaults to `latest`, which is the root cause of the selector breakage: an upstream UI change silently breaks every workspace using this module. Worth deciding separately whether to pin a known-good default. Tracked in coder/registry#1044.

Closes coder/registry#20<br>Closes coder/registry#1044


matifali:
@DevelopmentCats @35C4n0r this is ready for review when you get a chance. Branch updated with main, all checks green.

It closes #20 and #1044, and it's verified end to end on a real AWS Windows workspace: the account password round-trips exactly, and the session auto-connects with no clicks.

For context, #1036 fixed the same escaping bug in `amazon-dcv-windows` using the same approach, so this is consistent with what already shipped there.

35C4n0r (approved):
Image
The JS script works as expected, and the changes in main.tf + script LGTM.

PR #1036, opened 2026-07-30
fix(amazon-dcv-windows): escape special characters in DCV credentials

### LINKED ISSUE #1035: bug: special characters in amazon-dcv-windows admin_password break DCV login
### Current Behavior

The `amazon-dcv-windows` module interpolates `admin_password` into two places without escaping it for the target syntax, so passwords containing common special characters silently become a different password.

**1. PowerShell double-quoted string** — [`install-dcv.ps1:2`](<https://github.com/coder/registry/blob/main/registry/coder/modules/amazon-dcv-windows/install-dcv.ps1#L2>)

```powershell
$adminPassword = "${admin_password}"
```

`$`, backtick, and `"` are all interpreted inside a PowerShell double-quoted string. `$` starts a variable reference (usually expanding to empty), a backtick escapes the next character, and a `"` terminates the string. The local admin account then gets a password that differs from the one Coder shows, or the script fails to parse.

**2. Unencoded URL query parameter** — [`main.tf:57`](<https://github.com/coder/registry/blob/main/registry/coder/modules/amazon-dcv-windows/main.tf#L57>)

```tf
url = "https://localhost:${var.port}${local.web_url_path}?username=${local.admin_username}&password=${var.admin_password}"
```

The credentials go straight into the query string with no `urlencode`. An `&` truncates the password, `#` drops everything after it, `+` decodes as a space, and a bare `%` produces an invalid escape.

Either path ends in a failed DCV login, and whether a user hits it depends on which characters their generated password happens to contain, so it presents as a random, hard-to-reproduce failure.

### Expected Behavior

The password reaching the DCV login and the password set on the Windows account are both byte-identical to the value shown in the Coder UI, for any character.

### Steps to Reproduce

1. Add the `amazon-dcv-windows` module to a Windows template with `admin_password = "Pa$$w&rd`"1!"`(or any`random_password`that happens to include`$`, `&``` , ``  ``` ``, `"`, or `#`).
2. Start the workspace and open the **Web DCV** app.
3. Observe the login fails, and that the password in the app URL is truncated or altered compared to the workspace UI.

### Desired Solution

Encode each value for its destination, the same approach used for `windows-rdp` in coder/registry#1034:

| Destination | Encoding |
| -- | -- |
| PowerShell script | single-quoted string with `'` doubled |
| App URL query | `urlencode()` on both username and password |

A test asserting the rendered script and app URL round-trip a password containing ``` \ " ' ``  ``` `` $ & < > | # % +` would keep this from regressing.

> [!NOTE]  <br>Same class of bug as coder/registry#20, different module, so it needs its own fix and release.

Related to coder/registry#20  <br>Follow-up for coder/registry#1034

Edd88-pixel:
## Why

`admin_password` crossed two syntax boundaries without destination-specific escaping:

* the rendered PowerShell script used a double-quoted string, so `$`, backticks, and `"` could alter the password or break parsing;
* the Web DCV query string interpolated credentials directly, so characters such as `&`, `#`, `%`, and `+` could change parameter structure or decoded values.

## Changes

* Render the administrator password as a PowerShell single-quoted string and double literal apostrophes.
* Apply `urlencode()` separately to the Web DCV username and password query parameters.
* Add Terraform and TypeScript regression coverage for default credentials, the existing required-variable contract, and exact round-trips of `` \ " ' ` $ & < > | # % + ``.
* Bump `amazon-dcv-windows` from `1.1.1` to `1.1.2` with the repository version-bump script.

## Problem

The Windows account password and the password decoded from the DCV app URL could differ from the Terraform input because PowerShell interpolation and URL query parsing interpreted special characters.

## Fix

The PowerShell boundary now uses a single-quoted literal with `'` escaped as `''`. The URL boundary encodes `username` and `password` individually exactly once.

## Testing & Validation

* `bun test --timeout 30000 registry/coder/modules/amazon-dcv-windows/main.test.ts` — 5 passed.
* `terraform init -upgrade` — passed.
* `terraform validate` — passed.
* `terraform test -verbose` — 2 passed.
* `./scripts/terraform_validate.sh` with the changed-module CI inputs — passed.
* `./scripts/terraform_test_all.sh` with the changed-module CI inputs — 2 passed.
* `./scripts/ts_test_auto.sh` with the changed-module CI inputs — 5 passed.
* Targeted Prettier checks, `terraform fmt -check`, and `bun x tsc --noEmit` — passed.
* `bun run fmt:ci` — the touched files passed after targeted formatting; the repository-wide command still reports pre-existing formatting findings outside this module and a Windows plugin panic.
* The `bun run tftest` and `bun run tstest` wrappers cannot launch their `.sh` scripts under Bun on Windows; the same scripts passed directly through Git Bash as listed above.

## Regression coverage

The rendered Terraform state is inspected to verify that PowerShell `''` decodes back to `'` and that `URL.searchParams` restores the exact original username and password, including `` \ " ' ` $ & < > | # % + ``. Test failure messages do not print credential values.

## Type of Change

- [X] Bug fix

## Module Information

**Path:** `registry/coder/modules/amazon-dcv-windows`  <br>**New version:** `v1.1.2`  <br>**Breaking change:** \[ \] Yes \[x\] No

Closes coder/registry#1035


matifali:
@Edd88-pixel, have you tested the module end-to-end in a real Windows template on AWS?

Edd88-pixel:
Yes  I just provisioned a real Windows instance on AWS and tested the module E2E. The workspace provisioned successfully, and I was able to log in through Web DCV with the configured administrator credentials, including special characters. Everything worked as expected.

matifali (approved):

DevelopmentCats (approved):

PR #1038, opened 2026-07-31
refactor: fileBrowser script can handle multiple runs

### LINKED ISSUE #932: bug(file browser): Timeout error, database lock
When a workspace is up for more than a day or two, and the coder agent restarts, the file browser agent will fail with an error:

```
🛠️  Configuring filebrowser 
DB_PATH: .config/filebrowser.db
2026/06/24 06:10:45 No config file used
2026/06/24 06:10:45 Using database: /home/coder/.config/filebrowser.db
Error: timeout
```
I believe what's happening is the following:
- The coder agent and file browser are alive and well on the workspace
- the coder server triggers an agent restart on the workspace
- the agent startup script tries to start filebrowser with the previous process still running
- Timeout error because the previous process has a lock on `home/coder/.config/filebrowser.db`

Not sure if I've misconfigured something, or if the file browser startup script needs to be tweaked to check for an existing process.

Harsh9485:
## Description

Closes : #932 

I added a check to see if a filebrowser process is already running, if a process found, the script verifies whather the `FB_DATABASE` environment variable matches the current database path, if it matches, it mean file Browser is already configured and running for same database, so there's no need to configure it again. the script exits early instead of running the configuration a second time.   

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/filebrower`  
**New version:** `v1.1.6`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Screenshots
Screenshot 2026-07-30 193128


Harsh9485:
Hey @DevelopmentCats, could you review the PR when you have some time? It’s been open for quite a while now. I’d really appreciate it if you could take a look when you get a chance.

PR #1046, opened 2026-08-11
fix(.github/workflows): detect PR-changed modules via the pulls files API

bpmct:
Fixes the Module Scorecard Check commenting on modules a PR never touched, reported on #1034 where a `windows-rdp` PR got a scorecard comparison for `coder/git-clone`.

Linear: [REG-74](https://linear.app/codercom/issue/REG-74/unrelated-module-scorecards-on-pr-commnets)

## Root cause

The workflow detected changed modules with `git diff "${BASE_SHA}"...HEAD`, where `BASE_SHA` is `github.event.pull_request.base.sha` and HEAD is GitHub's test merge commit (PR head merged into **current** main). `base.sha` is the base tip from when the PR was opened, so on PRs whose base has since moved, the diff includes every module merged to main after the PR branched. #1034 was opened July 29; `git-clone` changed on main afterwards (#1037), so a later synchronize run scored it, and score noise made it look like a regression.

## Change

- Determine changed modules from `GET /repos/{owner}/{repo}/pulls/{n}/files` (paginated) instead of git plumbing. The PR files list is the source of truth for what a PR changes, regardless of how stale the branch is.
- Drop `fetch-depth: 0` from checkout; it existed only for the diff.
- Guard the `grep` so a PR with no module files doesn't fail the step under `pipefail`.

## Validation

Against #1034 directly: the files API returns only `windows-rdp` paths, while the old `base.sha` diff on the same PR spans dozens of unrelated commits including the `git-clone` change.

🤖 Generated with [Coder Agents](https://coder.com/docs/ai-coder/agents)

### DIFF
--- a/.github/workflows/module-scorecard-check.yaml
+++ b/.github/workflows/module-scorecard-check.yaml
@@ -43,17 +43,15 @@ jobs:
 
       - name: Determine changed modules
         id: changed
-        env:
-          GH_TOKEN: ${{ github.token }}
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-        # The PR files API is the source of truth for what a PR changes.
-        # Diffing against github.event.pull_request.base.sha is wrong: that
-        # SHA is the base tip when the PR was opened, while the checkout is
-        # the PR head merged into current main, so the diff picks up every
+        # The checkout is GitHub's test merge commit (PR head merged into
+        # current main), so HEAD^1 is the exact main tip this PR is applied
+        # to and the diff is precisely what the PR changes. Diffing against
+        # github.event.pull_request.base.sha is wrong: that SHA is the base
+        # tip from when the PR was opened, so on stale PRs it picks up every
         # module merged to main since the PR branched and scores unrelated
-        # modules (coder/registry REG-74).
+        # modules (REG-74).
         run: |
-          MODULES=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename' | { grep -oP '^registry/coder/modules/\K[^/]+' || true; } | sort -u | paste -sd, -)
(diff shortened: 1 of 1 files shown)


.github/workflows/module-scorecard-check.yaml
```diff
@@ -46,9 +44,16 @@ jobs:
...
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        # The PR files API is the source of truth for what a PR changes.
+        # Diffing against github.event.pull_request.base.sha is wrong: that
+        # SHA is the base tip when the PR was opened, while the checkout is
+        # the PR head merged into current main, so the diff picks up every
+        # module merged to main since the PR branched and scores unrelated
+        # modules (coder/registry REG-74).
         run: |
-          MODULES=$(git diff --name-only "${BASE_SHA}"...HEAD | grep -oP '^registry/coder/modules/\K[^/]+' | sort -u | paste -sd, -)
+          MODULES=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" --paginate --jq '.[].filename' | { grep -oP '^registry/coder/modules/\K[^/]+' || true; } | sort -u | paste -sd, -)
```
matifali: nit: I would discourage relying on gh api given its uptime. Is there an equivalent git command to do the same?

bpmct: Good call. Switched to a pure git approach that keeps the fix: the checkout is GitHub's test merge commit, so `HEAD^1` is the exact main tip the PR is applied to, and `git diff --name-only HEAD^1 HEAD` is precisely the PR's changes with no staleness (needs `fetch-depth: 2` for the parent). No API call in the detection step anymore.

matifali (approved):
Lets test it 🤞🏼

Issue #308, opened 2025-08-08
Epic: Migrate all modules from Bun tests to Terraform tests


matifali:
This epic tracks migrating module tests from Bun-based tests to Terraform native testing (`terraform test`).

- [x] Bootstrap Terraform test runner and docs
- [x] Migrate `coder/zed`
- [x] Migrate `coder/jetbrains`
- [x] Migrate `coder/code-server`
- [ ] Migrate `coder/agentapi`
- [ ] Migrate `coder/devcontainers-cli`
- [ ] Migrate `coder/kiro`
- [ ] Migrate `coder/jetbrains-fleet`
- [ ] Migrate `coder/jetbrains-gateway`
- [ ] Migrate `coder/vscode-web`
- [ ] Migrate `coder/vscode-desktop`
- [ ] Migrate `coder/vault-token`
- [ ] Migrate `coder/vault-jwt`
- [ ] Migrate `coder/vault-github`
- [ ] Migrate `coder/slackme`
- [ ] Migrate `coder/personalize`
- [ ] Migrate `coder/kasmvnc`
- [ ] Migrate `coder/jupyterlab`
- [ ] Migrate `coder/jfrog-token`
- [ ] Migrate `coder/jfrog-oauth`
- [ ] Migrate `coder/github-upload-public-key`
- [ ] Migrate `coder/git-config`
- [ ] Migrate `coder/git-clone`
- [ ] Migrate `coder/gcp-region`
- [ ] Migrate `coder/fly-region`
- [ ] Migrate `coder/dotfiles`
- [ ] Migrate `coder/cursor`
- [ ] Migrate `coder/coder-login`
- [ ] Migrate `coder/azure-region`
- [ ] Migrate `coder/aws-region`
- [ ] Migrate `anomaly/tmux`
- [ ] Migrate `coder-labs/gemini`
- [ ] Migrate `whizus/exoscale-instance-type`
- [ ] Migrate `whizus/exoscale-zone`
- [ ] Migrate `thezoker/nodejs`

Notes
- Remove corresponding `main.test.ts` when a module gets migrated.
- Prefer `plan` runs for structural assertions; use `apply` only when outputs are required.
- Keep Bun harness until all modules are migrated, then remove Bun deps and tooling.

matifali:
@DevelopmentCats I think we can simply migrate away from Bun tests as terraform doesn't provide a great way to run tests that execute scripts. 

It is possible to use docker provider and run tests in containers if we want to stay exclusively on terraform

PR #1047, opened 2026-08-11
feat(devcontainers-cli): support pinned and mirrored installs

Edd88-pixel:
## Why

The module always installed the latest `@devcontainers/cli` release from the package manager's configured public source. That made reproducible installs and restricted-network workspaces difficult, while global installs could depend on elevated permissions.

## Changes

- add validated inputs for an npm version or dist-tag and an optional npm-compatible registry URL
- install into Coder's agent-managed binary directory without `sudo`
- preserve existing package-manager detection and skip installation when `devcontainer` is already available
- propagate package-manager failures and verify the installed command is on `PATH`
- document pinned, mirrored, and air-gapped usage
- add Terraform tests, deterministic npm/pnpm/Yarn runtime coverage, and a real pinned npm installation test

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [x] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/devcontainers-cli`  
**New version:** `v1.2.0`  

## Validation

- [x] `terraform test` — 4 passed
- [x] `bun test registry/coder/modules/devcontainers-cli/main.test.ts` — 9 passed
- [x] real npm installation of `@devcontainers/cli@0.80.0` in `node:22-alpine`, followed by `devcontainer --version`
- [x] `terraform validate`
- [x] repository ShellCheck validation
- [x] repository README validation
- [x] targeted Prettier and Terraform formatting checks
- [x] `tsc --noEmit`

## Related Issues

Related to #427 and #308.


Edd88-pixel:
@matifali could you review this please ?

Edd88-pixel:
@DevelopmentCats Can you retry ?

registry/coder/modules/devcontainers-cli/run.sh
```diff
@@ -1,14 +1,33 @@
 #!/usr/bin/env sh
+# shellcheck shell=sh
+
+# shellcheck disable=SC2016 # Terraform replaces this literal placeholder.
```
matifali: I wonder why we had to use this?<br>If you do not want terraform to replace a literal. use `$${LITERAL`} or `$LITERAL`

registry/coder/modules/devcontainers-cli/run.sh
```diff
@@ -29,34 +48,47 @@ else
   exit 1
 fi
 
+INSTALL_PREFIX=$(dirname "$CODER_SCRIPT_BIN_DIR")
```
matifali, line 48: Should we allow provding an input to control this?

registry/coder/modules/devcontainers-cli/run.sh
Edd88-pixel: These placeholders are intentionally replaced by templatefile; the base64 encoding keeps the rendered shell values safe. The SC2016 suppressions are only needed because I used single quotes. 

We need Terraform’s templatefile to substitute the value while rendering the script, so using $${LITERAL} would leave the required value unset.

registry/coder/modules/devcontainers-cli/run.sh
Edd88-pixel: Do you mean allowing callers to provide a custom install prefix? I currently derive it from CODER_SCRIPT_BIN_DIR so npm/Yarn install into the Coder-managed bin directory already on PATH, while pnpm uses it as PNPM_HOME. I’d prefer to keep that fixed unless there is a specific use case, but I’m happy to adjust if that’s what you meant.

matifali: Yes. Sorry I forgot to complete what I was typing. Its not a blocker

registry/coder/modules/devcontainers-cli/run.sh
matifali: I understand. If you use `${LITERAL}` , Terraform's `templatefile()`will replace it but if you use `$${LITERAL}` or `$LITERAL` it wont.

Edd88-pixel: Updated it to keep ${LITERAL} for Terraform substitution, switched to double quotes, and removed the unnecessary SC2016 suppressions. 
I think it's for the best like that

matifali: Thanks. Try nto to use exceptions unless its very necessary. And always put a comment if you have to describing why.

PR #1050, opened 2026-08-17
feat: add devin-desktop module

matifali:
## Summary

Cognition rebranded the Windsurf Editor as Devin Desktop on June 2, 2026 (Cognition acquired Windsurf from Codeium in July 2025). This adds a `devin-desktop` module using the current name, without breaking the existing `windsurf` module.

## Backward compatibility approach

The Terraform module registry protocol has no rename/alias mechanism (`registry-server`'s own code comments confirm this: "the registry server has no alias mechanism for the old path"). Renaming `windsurf`'s directory/slug would break every template pinned to `registry.coder.com/coder/windsurf/coder`, so this PR is purely additive:

- **`windsurf` module: unchanged behavior.** Only its README gets a `[!IMPORTANT]` notice pointing to `devin-desktop`. It stays published and fully functional for existing templates.
- **New `devin-desktop` module**, matching the same pattern already used by every other desktop-IDE module in this registry (`cursor`, `zed`, `windsurf`, etc.): a thin wrapper around the shared `vscode-desktop-core` module, so none of the actual `coder_app`/deep-link logic is duplicated beyond what's already duplicated across that whole module family today.

I considered having `devin-desktop` delegate directly to the `windsurf` module (even less duplication), but that requires bumping `windsurf` to a new tagged version first, module registry composition resolves against already-published versions, and tags are only cut by a maintainer after merge (see `MAINTAINER.md`). That would force two sequential PRs with a manual tagging step in between for no real benefit given how small the `vscode-desktop-core` wiring is, so I went with the direct wrapper instead.

## Icon

Added a dedicated `devin.svg` (`.icons/devin.svg`), a monochrome vector of Cognition's Devin mark, adapted from the MIT-licensed [`@lobehub/icons-static-svg`](https://www.npmjs.com/package/@lobehub/icons-static-svg) package (`icons/devin.svg`), recolored to `#fff` to match this registry's existing monochrome icon convention (e.g. `windsurf.svg`). Named `devin`, not `devin-desktop`, matching how this registry already names icons/tags after the umbrella brand rather than the specific product surface (e.g. `claude-code` module uses `claude.svg`/`claude` tag). The module id/slug stays `devin-desktop` though, kept specific because Cognition's Devin family also includes Devin Local/Cloud/Review, a bare `devin` module name would be ambiguous between the desktop IDE launcher (this module) and a future CLI-agent-style module.

## ⚠️ Merge sequencing

This module defaults to opening Devin Desktop via a `devin://` deep link (placeholder, not independently verified against a live install). Opening `devin://` links requires `"devin:"` to be registered in Coder's `ALLOWED_EXTERNAL_APP_PROTOCOLS`, see [coder/coder#28214](https://github.com/coder/coder/pull/28214). **Do not tag/release this module until #28214 ships in a released Coder version**, otherwise the module's "Open" button is broken (unsubstituted session token) on any deployment running an older Coder. The `protocol` input can be overridden (e.g. back to `"windsurf"`) in the meantime if needed.

## Known gap (flagging for review)

Devin Desktop is documented by Cognition as backwards-compatible with Windsurf (same `~/.codeium/windsurf/` config path), but I haven't verified this hands-on against the actual app.

## Validation

- `terraform init -upgrade` and `terraform validate` pass against the live registry.
- `bun test registry/coder/modules/devin-desktop/main.test.ts`: 10/10 pass (mirrors `windsurf`'s test coverage, plus cases for overriding `slug`/`display_name`/`protocol`).
- `terraform fmt -check`, `bun run fmt`: clean for this module (repo-wide `terraform fmt -recursive` currently panics for unrelated pre-existing reasons; scoped `terraform fmt -check -diff` on this module's directory passes).
- `go build ./cmd/readmevalidation && ./readmevalidation`: passes (81 modules including the new one, all relative URLs and icon paths valid).

Related: [coder/coder#28205](https://github.com/coder/coder/pull/28205) (docs rebrand, now links to this module), [coder/coder.com#1014](https://github.com/coder/coder.com/pull/1014) (redirect), [coder/coder#28206](https://github.com/coder/coder/pull/28206) (Devin icon), [coder/coder#28214](https://github.com/coder/coder/pull/28214) (URI scheme allowlist, blocking).

> 🤖 This PR was created with the help of Coder Agents, and needs a human review.  🧑💻


registry/coder/modules/devin-desktop/main.tf
```diff
@@ -0,0 +1,117 @@
...
+  display_name       = "Devin Desktop MCP"
+  icon               = "/icon/devin.svg"
+  run_on_start       = true
+  start_blocks_login = false
+  script             = <<-EOT
+    #!/bin/sh
+    set -eu
+    mkdir -p "$HOME/.codeium/windsurf"
+    echo -n "${local.mcp_b64}" | base64 -d > "$HOME/.codeium/windsurf/mcp_config.json"
+    chmod 600 "$HOME/.codeium/windsurf/mcp_config.json"
```
matifali: research devin desktop docs to find the latest naming conventions.

registry/coder/modules/devin-desktop/README.md
```diff
@@ -0,0 +1,81 @@
+---
+display_name: Devin Desktop
+description: Add a one-click button to launch Devin Desktop (formerly Windsurf Editor)
+icon: ../../../../.icons/devin.svg
+verified: true
+tags: [ide, devin, windsurf, ai]
```
matifali: drop windsurf tag

registry/coder/modules/devin-desktop/README.md
```diff
@@ -0,0 +1,81 @@
...
+> Devin Desktop is Cognition's June 2, 2026 rebrand of the Windsurf Editor (itself a rebrand of Codeium's
+> original editor). It reuses Windsurf's `~/.codeium/windsurf/` MCP config path.
+
+> [!IMPORTANT]
+> This module defaults to opening Devin Desktop via a `devin://` deep link, a placeholder that has not
+> been independently verified against a live Devin Desktop install. Opening `devin://` links also requires
+> `"devin:"` to be registered in Coder's `ALLOWED_EXTERNAL_APP_PROTOCOLS`
+> ([coder/coder#28214](https://github.com/coder/coder/pull/28214)), so this module should not be tagged/released
+> until that ships in a released Coder version. Override the `protocol` input (e.g. back to `"windsurf"`) if
+> needed in the meantime.
```
matifali: drop these

registry/coder/modules/windsurf/README.md
matifali, line 1: this file would also need a patch update

registry/coder/modules/devin-desktop/main.tf
matifali: Checked docs.devin.ai. The FAQ confirms the \~/.codeium/ directory structure is not changing in this release (MCP config included), and the Cascade MCP docs still show the deep link format as windsurf://windsurf-mcp-registry?serverName=<server-name>. So \~/.codeium/windsurf/mcp_config.json is confirmed correct, not a guess. On protocol: docs still say windsurf://, but you tested devin:// directly and confirmed it works, so kept protocol default as devin (hands-on beats docs here). Dropped the hedging admonitions accordingly in the latest push.

matifali:
Note on versioning: `windsurf` is bumped to `1.3.2` in its README (matches its existing `v1.3.1` tag + patch, for the deprecation-notice change).

`devin-desktop` stays at `1.0.0`, it's a brand-new module with no prior tag. I initially added the `version:patch` label per review feedback, but the `version-bump.sh --ci` check treats any untagged module as needing README-version+1 every time it reruns (no tag to diff against), so it kept demanding another increment (1.0.1, then 1.0.2, ...) on every rerun. That's a pre-existing quirk in the script for brand-new modules sharing a PR with an existing one, not something to chase by incrementing further. Removed the label so that check doesn't gate this PR; `devin-desktop` will get its real `v1.0.0` tag via the normal post-merge `tag_release.sh` flow.

registry/coder/modules/devin-desktop/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  type        = string
+  description = "JSON-encoded string to configure MCP servers for Devin Desktop. When set, writes ~/.codeium/windsurf/mcp_config.json, the path Devin Desktop inherited from Windsurf."
+  default     = ""
+}
+
+variable "protocol" {
+  type        = string
+  description = "The URI protocol used to open Devin Desktop. Defaults to the devin:// scheme."
+  default     = "devin"
+}
```
matifali: I think we should drop this input. As devin:// works.

registry/coder/modules/devin-desktop/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  display_name       = "Devin Desktop MCP"
+  icon               = "/icon/devin.svg"
+  run_on_start       = true
+  start_blocks_login = false
+  script             = <<-EOT
+    #!/bin/sh
+    set -eu
+    mkdir -p "$HOME/.codeium/windsurf"
+    echo -n "${local.mcp_b64}" | base64 -d > "$HOME/.codeium/windsurf/mcp_config.json"
+    chmod 600 "$HOME/.codeium/windsurf/mcp_config.json"
```
matifali: Use the new location as in.
`$HOME/.config/devin/mcp_config.json`

phorcys420:
will review this later because it seems mostly AI generated so it will take more time to review thouroughly :)

DevelopmentCats (approved):
I tested this myself and everything works as expected with the devin `coder_app` and the mcp server config.

The only thing I notice @matifali is that you need Devin to be open already for it to actually open the remote connection to the coder workspace it seems like its an issue on their end that this only works when its running on the device you are launching the devin on from the workspace.

bpmct (approved):

matifali:
This is an issue with Both Cursor and Devin now as you need to be on the Editor screen for this to work.<br>@EhabY are we are of this limiation?

EhabY:
@matifali Yes we are aware, this is not an issue from our side, here's the Cursor thread: https://forum.cursor.com/t/installed-extensions-onuri-handler-is-not-invoked-for-custom-uri/167316/5

I do not have one for Devin but I'm guessing it's the same issue :/

matifali:
Yeah. Can we track that in coder/vscode-coder

EhabY:
Created https://github.com/coder/vscode-coder/issues/1087

PR #1056, opened 2026-08-18
fix(local-windows-rdp): escape special characters in credentials

### LINKED ISSUE #1043: bug: special characters in local-windows-rdp password break the setup script and app URL
### Current Behavior

The `local-windows-rdp` module interpolates `var.password` into two places without escaping it for the target syntax. With a password containing common special characters, the startup script fails to parse and the app URL is corrupted.

**1. PowerShell double-quoted string** — [`configure-rdp.ps1:97`](<https://github.com/coder/registry/blob/main/registry/coder/modules/local-windows-rdp/configure-rdp.ps1#L97>)

```powershell
$password = "${password}"
```

`$`, backtick, and `"` are interpreted inside a PowerShell double-quoted string. Observed on a live AWS Windows workspace with `password = "Te\st$Pa\"ss'w`ord&<>|1!"`, the whole `Configure RDP` script aborts before it does anything:

```
The string is missing the terminator: '.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString
```

The workspace then reports `start_error` and the admin password is never set.

**2. Unencoded URL query parameter** — [`main.tf:75`](<https://github.com/coder/registry/blob/main/registry/coder/modules/local-windows-rdp/main.tf#L75>)

```tf
url = "coder://${local.server_name}/v0/open/ws/${data.coder_workspace.me.name}/agent/${var.agent_name}/rdp?username=${var.username}&password=${var.password}"
```

The credentials go straight into the query string with no `urlencode`. The rendered app URL on the same workspace was:

```
coder://dev.coder.com/v0/open/ws/reg37-rdp/agent/main/rdp?username=Administrator&password=Te\st$Pa"ss'w`ord&<>|1!
```

The password truncates at the first `&`, so Coder Desktop receives `Te\st$Pa"ss'w`ord`. A `#`would drop everything after it,`+`decodes as a space, and a bare`%` is an invalid escape.

### Expected Behavior

The password set on the Windows account and the password handed to Coder Desktop are both byte-identical to `var.password`, for any character.

### Steps to Reproduce

1. Use the `local-windows-rdp` module on a Windows workspace with `password = "Te\st$Pa\"ss'w`ord&<>|1!"`.
2. Start the workspace.
3. The `Configure RDP` script fails with the parser error above, and the `Open RDP` app URL is truncated at the `&`.

### Desired Solution

Encode each value for its destination, the same approach used for `windows-rdp` in coder/registry#1034:

| Destination | Encoding |
| -- | -- |
| PowerShell script | single-quoted string with `'` doubled |
| App URL query | `urlencode()` on both username and password |

A test asserting the rendered script and app URL round-trip a password containing ``` \ " ' ``  ``` `` $ & < > | # % +` would keep this from regressing.

> [!NOTE]
> Third module with this bug class, after coder/registry#20 (`windows-rdp`) and coder/registry#1035 (`amazon-dcv-windows`). Found while dogfooding the fix in coder/registry#1034 on a real AWS Windows workspace.

Related to coder/registry#20, coder/registry#1035  <br>Follow-up for coder/registry#1034

matifali:
## Problem

The password was interpolated into two places without escaping it for the target syntax.

**1. PowerShell double-quoted string** — `configure-rdp.ps1`

```powershell
$password = "${password}"
```

`$`, backtick, and `"` are all interpreted there. Reproduced on a live AWS Windows workspace with `password = "Te\st$Pa\"ss'w`ord&<>|1!"`: the entire `Configure RDP`script aborts before doing anything, and the workspace reports`start_error\`.

```
The string is missing the terminator: '.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString
```

**2. Unencoded URL query parameter** — `main.tf`

The same value went into the `coder://` app URL raw, so the rendered URL on that workspace was:

```
coder://dev.coder.com/v0/open/ws/reg37-rdp/agent/main/rdp?username=Administrator&password=Te\st$Pa"ss'w`ord&<>|1!
```

Coder Desktop receives everything up to the first `&`. A `#` drops the remainder, `+` decodes as a space, and a bare `%` is an invalid escape.

## Fix

| Destination | Encoding |
| -- | -- |
| PowerShell script | single-quoted string, `'` doubled |
| App URL query | `urlencode()` on both username and password |

Same approach as coder/registry#1036 for `amazon-dcv-windows` and coder/registry#1034 for `windows-rdp`.

## Changes

* `configure-rdp.ps1`: pass credentials as PowerShell single-quoted strings
* `main.tf`: `urlencode()` both credentials in the app URL
* add `local-windows-rdp.tftest.hcl` covering a password with ``` \ " ' ``  ``` \`\` $ & < > | # % +\`
* update 4 assertions in `main.test.ts` that were pinning the old behavior, for example `password=CustomPass123!` unencoded

Module bumped to `1.0.5`.

## Validation

The new Terraform tests fail without the source fix and pass with it, so they are not vacuous:

```
run "plan_with_defaults"...           fail
  condition = strcontains(resource.coder_script.rdp_setup.script, "$password = 'coderRDP!'")
run "plan_with_special_characters"... fail
```

With the fix: `2 passed, 0 failed`, `bun test main.test.ts` 10/10, prettier and `terraform fmt` clean.

> [!NOTE]
> Third and last module in this bug class, after coder/registry#20 / coder/registry#1034 (`windows-rdp`) and coder/registry#1035 / coder/registry#1036 (`amazon-dcv-windows`). Found while dogfooding coder/registry#1034 on a real AWS Windows workspace, where this module's script was the one that actually crashed.

Closes coder/registry#1043


matifali:
On the scorecard regression: the Credential Hygiene drop is scoring noise, not something this PR introduced.

The only README change in the original diff was the version bump, `1.0.4` → `1.0.5`, on three lines. The criterion the scorecard cites, the inline `coderRDP!` default in an example, is identical on `main`:

```
$ git show origin/main:registry/coder/modules/local-windows-rdp/README.md | grep -n coderRDP
54:Uses default credentials (Username: `Administrator`, Password: `coderRDP!`):
```

Presentation moved 17.5 → 20 in the same run with no README content change either, which is the same variance in the other direction. The footer notes these are advisory language-model scores.

That said, the underlying point about the non-hardcoded auth path is fair, so I've added a `random_password` example in 226133f. It's on-theme for this PR: generating a password is exactly the case that used to break, since a generated value is likely to contain `\`, `&`, or `$`.

The remaining criteria I'm deliberately not addressing here:

- **Visual preview (0/5)**: needs a screenshot or GIF, unrelated to a credential escaping fix.
- **Inline default in the example**: documenting the module's actual default is legitimate, and removing it would make the README less useful.

DevelopmentCats (approved):

PR #1057, opened 2026-08-18
fix(registry/coder/modules/vscode-web): install extensions when reusing a cached copy

35C4n0r:
## Description

Separate "skip the download" from "skip extensions" so a cached or pre-installed VS Code Web still installs extensions:

- **Offline**: unchanged. Runs the pre-baked copy and exits, or fails when no copy is present.
- **Cached (`use_cached`)**: reuse the existing copy, skip only the download, then install extensions and start the server like a fresh install.
- **Default**: unchanged (always downloads).

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/vscode-web`<br>
**New version:** `v1.6.2`<br>

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

REG-63

---

🤖 Generated by Coder Agents on behalf of @35C4n0r.


DevelopmentCats:
> Visual preview | 5 | 2 | README includes a GIF (vscode-web.gif) showing VS Code Web with GitHub Copilot and live-share, but the path ../../.images/vscode-web.gif appears broken (relative path goes outside module directory). Partial credit for attempting visual preview with unclear accessibility.

This is a false positive I am going to address this seperately but other than that this looks good.

DevelopmentCats (approved):

DevelopmentCats:
Tested this on a live deployment the fix works. `use_cached` now reuses the copy, skips only the download, and still installs both the `extensions` list and the `auto_install_extensions` recommendations. Confirmed by wiping `extensions_dir` between restarts and watching it repopulate. `offline` and the default download path are unchanged.

One small thing in `run.sh`:

**1. The "Found a copy" message now prints on the default path, right before it downloads anyway**

Hoisting the `echo` out of the mode check means it fires whenever a copy exists, including when neither `offline` nor `use_cached` is set:

```bash
if [ -f "$VSCODE_WEB" ]; then
  echo "🥳 Found a copy of VS Code Web"
  if [ "${OFFLINE}" = true ]; then
```

Before this PR the echo lived inside `if [ "${OFFLINE}" = true ] || [ "${USE_CACHED}" = true ]`, so it only appeared when the copy was actually going to be used.

Anyone with a persistent `install_prefix` who hasn't opted into `use_cached` now gets this on every restart:

```
🥳 Found a copy of VS Code Web
Installing Microsoft Visual Studio Code Server!
VS Code Web commit id version 110a328ea54b42367b803ec53ee0bf52ef26b419.
VS Code Web has been installed.
```

Which reads like the cache was used when it wasn't. Purely cosmetic, but moving the echo into the two branches that act on it would be accurate.

PR #525, opened 2025-11-07
add oracle oci-vm template

aniskhalfallah:
## Description

Add oci-vm template to coder

## Type of Change

- [ ] New module
- [x] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Template Information

**Path:** `registry/anis/templates/oci-vm`

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Code formatted (`bun fmt`)
- [x] Changes tested locally

https://github.com/user-attachments/assets/2164a766-7838-415e-bbcc-04ba58322b53

## Related Issues

Closes #201 
/claim #201

PR #1058, opened 2026-08-18
fix: repair oci-vm template home volume mount, region default, and ssh debug access

bpmct:
## Description

While testing @aniskhalfallah's OCI VM template (#525) end-to-end against a real OCI tenancy, I found a few things that don't actually work once you get past the reviews:

- The home volume never mounts. `disk_setup`/`mounts` run early in cloud-init's boot sequence, before the OCI volume attachment (a separate, async API call) has actually shown up as a device. Confirmed via the instance's serial console: `Device /dev/sdb did not exist and was not created with a udevadm settle`. `/home/<user>` silently ends up on the ephemeral root disk instead, so it's wiped on every stop despite the template (and README) claiming it persists.
- `region` defaults to `eu-marseille-1`, which most tenancies aren't subscribed to (including a fresh trial account, per the review thread, and my own tenancy's home region). Since the AD/image data sources aren't gated by `start_count`, this makes `coder templates push` itself fail with an auth-looking 401 before anyone even gets to create a workspace.
- `ssh_public_key` is documented ("SSH public key for debugging access") but never wired into cloud-init, so there's no way to get into a workspace to debug it.

This PR builds on #525's `main.tf`/README/cloud-init with:
- A `runcmd` script that waits for the volume device to show up, formats it only if it isn't already formatted (matched by filesystem label, so repeated stop/start cycles don't wipe it), and mounts it via `/etc/fstab`. Also fixes a subtler bug this introduced: ext4 labels are capped at 16 bytes, but the existing `home_volume_label` local is (correctly, for OCI's own 32-char display name limit) longer than that, so `mkfs` was silently truncating it and later lookups by the untruncated label never matched. Added a dedicated 16-byte `home_fs_label`.
- `region` now defaults to `us-ashburn-1` instead of `eu-marseille-1`. Still not universal (no single default can be, since OCI only auto-subscribes your home region), but it's one of OCI's two original commercial regions and a much more common default than a specialty region — documented the caveat clearly either way.
- `ssh_public_key` now actually lands in `~/.ssh/authorized_keys`. Had to provision it from inside the mount script rather than through cloud-init's normal `ssh_authorized_keys`, since that runs before the home volume mount and would otherwise get shadowed by it.

Tested all of the above against a real OCI tenancy (instance + volume + attachment, connected agent, `df`/`lsblk` showing `/home` on the block volume, `ssh <user>@<ip>` working with the debug key).

Opening as a draft since this is meant as a starting point — happy to fold it into #525 directly, split it up, or adjust approach based on what @aniskhalfallah and @DevelopmentCats prefer.

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Template Information

**Path:** `registry/anis/templates/oci-vm`

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

Related to #525


bpmct:
Redirecting this to target @aniskhalfallah's branch directly instead of main: https://github.com/aniskhalfallah/registry/pull/1

PR #1060, opened 2026-08-19
feat(gojnimer6553): add herdr module

gojnimer6553:
## Description

Adds a `herdr` module: runs [Herdr](https://herdr.dev) — a terminal/agent-session manager that detects and drives coding agents (Claude Code, Codex, OpenCode, and others) in persistent background panes — inside a Coder workspace, and optionally installs Herdr [plugins](https://herdr.dev/docs/plugins/) non-interactively via `herdr plugin install <spec> --yes`.

Herdr has no documented headless/daemon startup mode, so this module gives it one via a detached `tmux` session, so the server keeps running independent of the `coder_script` that launched it.

Key design points:
- `plugins` (opt-in `list(string)`, default `[]`) — each entry is an `owner/repo` spec installed non-interactively on every start. Installing a plugin only registers it; most plugins still need their own one-time interactive setup from a workspace terminal afterwards (documented in the README).
- `app_port` (optional) — exposes a plugin's own local web UI as a `coder_app` tile.
- `post_start_script` (optional) — runs after Herdr and all configured plugins are up, so callers can work around plugin-specific gaps (e.g. a plugin's setup wizard assuming infrastructure the workspace doesn't have) without the module itself needing to know about any specific plugin.

## Type of Change

- [x] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/gojnimer6553/modules/herdr`
**New version:** `v1.0.0`

## Testing & Validation

- [x] Tests pass (`terraform test`: 14/14, `bun test main.test.ts`: 10/10 against real Docker containers)
- [x] Code formatted (`bun run fmt`, shellcheck clean)
- [x] Changes tested locally, including against the real `herdr` CLI and a real Herdr plugin (`AltanS/collie`) in a live Coder workspace template

## Related Issues

None


DevelopmentCats:
`scripts/start.sh.tftpl` — `wait_for_herdr` can't ever fail, so `✅ Herdr is up.` prints whether or not it is. `herdr status` exits 0 even when the server is stopped (checked on 0.8.2: `herdr server stop` → `status: not running`, `exit=0`), so the loop returns on the first iteration.

Forced a Herdr that crashes a few seconds after launch and ran this script against it:

```
🚀 Starting Herdr in the background (tmux session 'herdr')...
✅ Herdr is up.
START SCRIPT EXIT=0
→ 0 herdr server procs, status: not running
```

Green workspace, no Herdr. Worth caring about because the module always installs latest, so one bad upstream release fails this way silently.

```bash
if herdr status 2>/dev/null | grep -q 'status: running'; then
```

Dropping the check entirely is also fine `code-server` and others just background the process and let the app healthcheck report it. But as written it's a check that only ever says yes.

If you do keep it: `HERDR_SESSION` is currently only passed to `tmux new-session`, so once the check really talks to the server it'll hit the default session's socket instead of yours. Export it and both agree.

DevelopmentCats:
`main.tf`, `variable "icon"` — the fork-branch URL breaks for everyone once this branch is deleted, and no other module hotlinks raw.githubusercontent.

Your comment on the variable already has the diagnosis right: herdr isn't in Coder's built-in icon bundle, so `/icon/herdr.svg` 404s today. The normal fix is to add it via a PR `site/static/icon/herdr.svg` into coder/coder (your `.icons/herdr.svg` can go straight in, next to `claude.svg`, `goose.svg`, `opencode.svg`), then set this default to `/icon/herdr.svg` like `claude-code` does. Once both merge it just works.

One tweak for that PR: the SVG has a solid `<rect fill="#d9dad8"/>` background, so it'll show as a light square in dark mode. Most bundle icons are transparent.

DevelopmentCats:
@gojnimer6553 I left a few comments but thank you for the contribution it looks good otherwise!

PR #1063, opened 2026-08-19
fix(claude-code): propagate installation failures

Edd88-pixel:
## Why

Claude Code installation failures could be logged while the module continued and reported success. The skip-install path could also complete without a usable pre-installed binary, leaving a workspace apparently configured while Claude Code was unavailable.

## Changes

- download the official installer to a temporary file with bounded retries and timeouts before execution
- validate the installer syntax, propagate download and installation failures, and verify the installed binary
- require a usable pre-installed binary when installation is disabled and expose its resolved path to later configuration steps
- cover installation, binary validation, authentication, MCP, telemetry, and provider scenarios with deterministic module tests
- bump the module patch version to 5.4.1

## Validation

- E2E container tests completed successfully: bun test main.test.ts — 26 passed, 0 failed (196 assertions)
- Terraform tests: terraform test — 27 passed, 0 failed
- Terraform configuration: terraform validate — passed
- TypeScript: bun x tsc --noEmit — passed
- Formatting: targeted Prettier checks and terraform fmt -check — passed
- Shell: ShellCheck on the rendered installer template — passed


registry/coder/modules/claude-code/main.test.ts
```diff
@@ -208,12 +252,108 @@ describe("claude-code", async () => {
...
+      "/home/coder/.coder-modules/coder/claude-code/logs/install.log",
+    );
+    expect(installLog).toContain("Claude Code binary was not found.");
+  });
+
+  test("resolved-pre-installed-binary-is-available-to-later-steps", async () => {
+    const { id, coderEnvVars, scripts } = await setup({
+      skipClaudeMock: true,
+      moduleVariables: {
+        claude_binary_path: "/opt/claude/bin",
```
35C4n0r, line 334: @Edd88-pixel the test `resolved-pre-installed-binary-is-available-to-later-steps` here is misleading, the mock claude binary is written to `~/.local/bin/claude` and the path you are passing here is `"/opt/claude/bin"`, so instead of using the preinstalled binary path the code flow from `resolve_claude_binary` uses the fallback `"$HOME/.local/bin/claude"`

35C4n0r (changes requested):

registry/coder/modules/claude-code/scripts/install.sh.tftpl
```diff
@@ -76,50 +76,77 @@ function add_path_to_shell_profiles() {
...
+function configure_claude_binary() {
+  local action="$1"
+  local CLAUDE_BIN
+  local CLAUDE_VERSION
+
+  CLAUDE_BIN=$(resolve_claude_binary)
+
+  if ! CLAUDE_VERSION=$("$${CLAUDE_BIN}" --version); then
+    echo "Claude Code binary could not be executed." >&2
+    return 1
```
35C4n0r, line 101: @DevelopmentCats @matifali do we want a hard failure here ?

matifali: I think yes as the primary purpose of this module is to install Claude Code and if its not on path. We should fail

registry/coder/modules/claude-code/main.test.ts
Edd88-pixel: Good catch , I'm working on it

35C4n0r (approved):

PR #1065, opened 2026-08-19
ci: allow maintainers to run the module scorecard on fork PRs via /scorecard

bpmct:
Fork PRs currently get no scorecard comment because they don't have access to secrets. This adds an `issue_comment` trigger: a maintainer comments `/scorecard` on a community PR to run the check on demand. Same-repo PRs keep the automatic `pull_request` behavior, unchanged.

The first version of this used `pull_request_target`, which zizmor rightly blocked. This approach avoids it: `issue_comment` runs in the base-repo context, the maintainer restriction (`OWNER`/`MEMBER`/`COLLABORATOR`) makes every fork-PR scoring an explicit human decision, and it doubles as cost control for the Anthropic API.

Discussions are not affected. PR mode only reads them for the baseline, and the token only has `discussions: read` anyway.

To keep secrets safe, the scoring scripts always run from the base repo. PR content is fetched via `refs/pull/N/merge` and only pulled in under `registry/coder/modules/`, read as plain text for the prompt, never executed.

🤖 Generated with Coder Agents on behalf of @bpmct

### DIFF
--- a/.github/workflows/module-scorecard-check.yaml
+++ b/.github/workflows/module-scorecard-check.yaml
@@ -4,8 +4,19 @@
 # regressions, celebrates improvements, and confirms unchanged scores.
 #
 # Never creates or updates discussions; the comparison is read-only against
-# GitHub and only writes the PR comment. Skipped for fork PRs because it
-# requires secrets.
+# GitHub and only writes the PR comment.
+#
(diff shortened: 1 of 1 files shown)


DevelopmentCats (approved):
LGTM

PR #1072, opened 2026-08-21
fix(github-upload-public-key): use ssh_signing_keys uri

v-atsai:
## Description

- The type of SSH key must be specified when uploading otherwise the upload key is not used to verify commits.
- Reference: https://docs.github.com/en/rest/users/ssh-signing-keys?apiVersion=2026-03-10

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/github-upload-public-key`  
**New version:** `v1.1.1`  
**Breaking change:**
- [ ] Yes
- [x] No

## Testing & Validation

- [ ] Tests pass (`bun test`)
- [ ] Changes tested locally

### DIFF
--- a/registry/coder/modules/github-upload-public-key/run.sh
+++ b/registry/coder/modules/github-upload-public-key/run.sh
@@ -68,7 +68,7 @@ GITHUB_KEYS_RESPONSE=$(
     -H "Accept: application/vnd.github+json" \
     -H "Authorization: Bearer $GITHUB_TOKEN" \
     -H "X-GitHub-Api-Version: 2022-11-28" \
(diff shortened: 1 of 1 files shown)


35C4n0r (changes requested):
@v-atsai adding a signing key should be an additional feature in the module, we don't want to replace the current behaviour of the module from adding authentication SSH key to signing SSH key.
I'd suggest to add a new optional variable where users can key_type auth(default)/signing/both, this would add signing as an additive option rather than a replacement.

PR #1082, opened 2026-08-27
chore: add bun check and check:* script namespace

phorcys420:
## What

Group the test/validate npm scripts under a consistent `check:*` namespace and add a single aggregate command.

| Before | After |
| --- | --- |
| `tftest` | `check:terraform` |
| `tstest` | `check:typescript` |
| `terraform-validate` | `check:validate` |
| `shellcheck` | `check:shellcheck` |
| — | `check` (runs all four) |

## Why `check` and not `test`

`bun test` is reserved by Bun's native test runner and never resolves a package.json `test` script, so an aggregate named `test` would be unreachable via `bun test`. The aggregate is named `check` and is invoked as `bun check`.

## CI

The four CI steps in `ci.yaml` now call `bun run check:*`. They're kept as separate steps so each check keeps its own annotations/visibility. The paths-filter and `set -u` ordering step are unchanged. Underlying `scripts/*.sh` filenames are unchanged; only the npm script names moved.

Docs referencing the old names were updated: `AGENTS.md`, `.agents/skills/coder-modules/SKILL.md`, `.agents/skills/coder-templates/SKILL.md`.

## Verification

- `bun check` runs all four in order, exit 0
- `bun x prettier --check` passes on all edited files

---
Generated with Coder Agents.

### DIFF
--- a/.agents/skills/coder-modules/SKILL.md
+++ b/.agents/skills/coder-modules/SKILL.md
@@ -329,12 +329,12 @@ Cleanup of `*.tfstate` files and `modules-test` Docker containers is handled aut
 | Task             | Command                                               | Scope      |
 | ---------------- | ----------------------------------------------------- | ---------- |
 | Format all       | `bun run fmt`                                         | Repo       |
-| Terraform tests  | `bun run tftest`                                      | Repo       |
-| TypeScript tests | `bun run tstest`                                      | Repo       |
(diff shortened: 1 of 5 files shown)


phorcys420:
main reason for this is so that i can just do `bun format && bun check`` before committing, and in the near future i wanna add commit hooks as well

matifali (approved):
Thank you. Its always a joy to see you contributing such QoL and DX enhancements.

PR #1092, opened 2026-09-02
feat(windows-rdp): add native client option

### LINKED ISSUE #1061: windows-rdp: Add option to open in native RDP desktop client
## Feature Request

The `windows-rdp` module (powered by Devolutions Gateway) currently only provides a web-based RDP client. It would be useful to have an option to launch the native RDP desktop client (e.g., Microsoft Remote Desktop, mstsc.exe) directly.

### Use Case

Some users prefer the native RDP experience for:
- Better performance and responsiveness
- Full clipboard integration
- Multi-monitor support
- Familiar keyboard shortcuts
- Integration with local RDP credential managers

### Proposed Implementation

Add a variable to optionally create a `coder_app` with `external = true` that generates an `.rdp` file or uses an `rdp://` URI scheme to launch the native client.

```tf
variable "enable_native_rdp" {
  type        = bool
  description = "Add an app button to open in the native RDP desktop client."
  default     = false
}
```

This could work similarly to how `local-windows-rdp` handles native RDP connections, but integrated into the main `windows-rdp` module as an optional feature.

### Related

- `coder/local-windows-rdp` module handles native RDP but is a separate module
- Consider whether these modules should be unified or if this should remain a separate concern

Edd88-pixel:
## Why

The `windows-rdp` module currently exposes only the browser-based  client. This change also provides a one-click native RDP experience through Coder Desktop.

Native RDP is enabled by default. Existing users who want to keep web-only access can set `enable_native_rdp = false`.

## Changes

- add an `RDP Desktop` app, enabled by default, using the existing Coder Desktop `coder://` RDP URI pattern
- add `agent_name`, required when native RDP is enabled, with cross-variable validation and a Terraform minimum version of 1.9
- allow users to opt out of the native app with `enable_native_rdp = false` while preserving the existing Web RDP app
- URL-encode the administrator username and password passed to the native client
- add a dedicated native RDP icon and a configurable tooltip explaining the Coder Desktop requirement
- add Terraform coverage for the disabled state, the required agent name, and the default-enabled state
- document the default behavior, web-only opt-out, Coder Desktop prerequisite, and `1.4.0` usage examples

## Validation

- `terraform fmt -check` — passed
- `terraform validate` — passed
- `terraform test` — 3 passed, 0 failed
- integrated-browser validation — one visible and enabled `RDP Desktop` link rendered with the expected `coder://.../agent/main/rdp` URI

## Description

Adds native RDP access through Coder Desktop by default while keeping the existing Web RDP app available. Consumers must provide `agent_name` when native RDP is enabled, or set `enable_native_rdp = false` for web-only access.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/windows-rdp`  
**New version:** `v1.4.0`  

Compatibility note: upgrading consumers must provide `agent_name` or explicitly disable native RDP. The module now requires Terraform 1.9 or newer.

## Testing & Validation

- [x] Tests pass (`terraform test` and `bun test`)
- [x] Code formatted (`terraform fmt -check`)
- [x] Changes tested locally

## Related Issues

Closes #1061


registry/coder/modules/windows-rdp/main.tf
```diff
@@ -53,6 +53,18 @@ variable "agent_id" {
...
+variable "agent_name" {
+  description = "The name of the Coder agent used for native RDP connections."
+  type        = string
+  default     = "main"
+}
+
+variable "enable_native_rdp" {
+  description = "Add an app button that opens the workspace in a native RDP client through Coder Desktop."
+  type        = bool
+  default     = false
```
matifali: lets default it to `true`

registry/coder/modules/windows-rdp/main.tf
```diff
@@ -125,6 +141,19 @@ resource "coder_app" "windows-rdp" {
   }
 }
 
+resource "coder_app" "native-rdp" {
+  count = var.enable_native_rdp ? 1 : 0
+
+  agent_id     = var.agent_id
+  display_name = "RDP Desktop"
+  slug         = "rdp-desktop"
+  icon         = "/icon/desktop.svg"
```
matifali: Is there another icon we can use? This may get confused with the existing Web RDP app icon.

registry/coder/modules/windows-rdp/README.md
```diff
@@ -10,12 +10,16 @@ tags: [windows, rdp, web, desktop]
 
 Enable Remote Desktop + a web based client on Windows workspaces, powered by [devolutions-gateway](https://github.com/Devolutions/devolutions-gateway).
 
+Set `enable_native_rdp = true` to add an **RDP Desktop** app that opens the
+workspace in a native RDP client through [Coder Desktop](https://coder.com/docs/user-guides/desktop).
+The native app uses the same administrator credentials as the web client.
```
matifali: Suggestion to make it true by default.

registry/coder/modules/windows-rdp/main.tf
```diff
@@ -53,6 +53,18 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
+variable "agent_name" {
+  description = "The name of the Coder agent used for native RDP connections."
+  type        = string
+  default     = "main"
+}
```
matifali, line 65: Can we make it cross-link to enable_native_rdp?
As it should only be required when tahst true.

registry/coder/modules/windows-rdp/main.tf
```diff
@@ -53,6 +53,18 @@ variable "agent_id" {
   description = "The ID of a Coder agent."
 }
 
+variable "agent_name" {
+  description = "The name of the Coder agent used for native RDP connections."
+  type        = string
+  default     = "main"
```
matifali:
```suggestion
```
Also, let's drop the default value, as we should not assume the `agent_name` for our customers.

registry/coder/modules/windows-rdp/main.tf
```diff
@@ -125,6 +141,19 @@ resource "coder_app" "windows-rdp" {
...
+
+  agent_id     = var.agent_id
+  display_name = "RDP Desktop"
+  slug         = "rdp-desktop"
+  icon         = "/icon/desktop.svg"
+  external     = true
+  order        = var.order
+  group        = var.group
+  url          = "coder://${regex("https?:\\/\\/([^\\/]+)", data.coder_workspace.me[0].access_url)[0]}/v0/open/ws/${data.coder_workspace.me[0].name}/agent/${var.agent_name}/rdp?username=${urlencode(var.admin_username)}&password=${urlencode(var.admin_password)}"
+}
```
matifali, line 171: Please also add [`tooltip`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/app#tooltip-1) with its value saying that this app requires Installing Coder Desktop.

See an example here: https://github.com/coder/registry/blob/878fcf2f29e60092fb0a450457149641deb44253/registry/coder/modules/jetbrains/main.tf#L62

But reformat it for Coder Desktop: https://coder.com/docs/user-guides/desktop

matifali (changes requested):

registry/coder/modules/windows-rdp/main.tf
Edd88-pixel: it's okay

registry/coder/modules/windows-rdp/main.tf
```diff
@@ -54,15 +54,21 @@ variable "agent_id" {
...
-  default     = "main"
+  default     = null
 }
 
 variable "enable_native_rdp" {
   description = "Add an app button that opens the workspace in a native RDP client through Coder Desktop."
   type        = bool
-  default     = false
+  default     = true
+}
```
matifali, line 71: We still need the [cross-validation condition](https://www.hashicorp.com/en/blog/terraform-1-9-enhances-input-variable-validations) for these variables.

Edd88-pixel: Good catch, thanks for the link,  I moved the check to a cross-variable validation on agent_name and updated the minimum Terraform version to 1.9.

DevelopmentCats:
I think the one main thing that I notice here @Edd88-pixel is that this isn't really an opt-in change as the PR describes its defaulted to true so its something that will be enforced for anyone using this new module version.

Edd88-pixel:
I’ll update the PR description to reflect that native RDP is enabled by default now.

matifali (approved):

DevelopmentCats (approved):

PR #1093, opened 2026-09-03
fix(git-clone): scan custom SSH ports

### LINKED ISSUE #1073: bug: `git-clone` did not parse and add port to `ssh-keyscan` command
## Issue
PR https://github.com/coder/registry/pull/942 add feature to add repo url to known host, but it only parse the host and did not parse the port when running `ssh-keyscan` command. This make the known host added for port 22 instead.

My use case is using self hosted gitea/forgejo, because port 22 already used by the host for ssh into the host machine I use port 2222 instead for the git ssh. The repo URL is something like this: 
```
ssh://git@forgejo.domain.tld:2222/LazyGeniusMan/test-repo.git
```

The current parsing is :
https://github.com/coder/registry/blob/6be121e4f6eabd2f248cb9c81953af4c95457634/registry/coder/modules/git-clone/run.sh#L56

The result for repo URL above is : `forgejo@domain.tld`

Then the value used in:
https://github.com/coder/registry/blob/6be121e4f6eabd2f248cb9c81953af4c95457634/registry/coder/modules/git-clone/run.sh#L64

## Proposed Fix
1. Parse the port from ssh repo url
2. Pass the `-p` arg to `ssh-keyscan` command

35C4n0r:
## Description

The git-clone module scanned SSH host keys on port 22 even when an `ssh://` repository URL specified a custom port, causing host-key verification to fail for self-hosted Git services on ports such as 2222.

Parse the host and port from `ssh://` URLs, use the port-specific `[host]:port` identity for known-host lookups, and pass the port to `ssh-keyscan`. Preserve scp-style URL behavior and add regression coverage for custom-port and existing-key cases.

## Type of Change

- [ ] New module
- [ ] New template
- [x] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**New version:** `v2.0.4`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

Fixes #1073
Fixes REG-93

---

Generated by Coder Agents for @35C4n0r.

### DIFF
--- a/registry/coder/modules/git-clone/README.md
+++ b/registry/coder/modules/git-clone/README.md
@@ -14,7 +14,7 @@ This module allows you to automatically clone a repository by URL and skip if it
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "2.0.3"
+  version  = "2.0.4"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
 }
@@ -28,7 +28,7 @@ module "git-clone" {
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "2.0.3"
+  version  = "2.0.4"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
   base_dir = "~/projects/coder"
@@ -43,7 +43,7 @@ To use with [Git Authentication](https://coder.com/docs/v2/latest/admin/git-prov
 module "git-clone" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/git-clone/coder"
-  version  = "2.0.3"
+  version  = "2.0.4"
   agent_id = coder_agent.example.id
   url      = "https://github.com/coder/coder"
 }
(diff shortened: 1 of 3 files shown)


DevelopmentCats:
@35C4n0r The TF Tests are failing so if you can resolve that we are good but other that that I tested and everything works fine

35C4n0r:
> @35C4n0r The TF Tests are failing so if you can resolve that we are good but other that that I tested and everything works fine

@DevelopmentCats transient error, re-ran the tests and everything passes now.

DevelopmentCats:
> > @35C4n0r The TF Tests are failing so if you can resolve that we are good but other that that I tested and everything works fine
> 
> @DevelopmentCats transient error, re-ran the tests and everything passes now.

Yeah I was looking at it and just came to the same conclusion approving now

DevelopmentCats (approved):

DevelopmentCats (changes requested):
Oh wait @35C4n0r You need to bump the version

35C4n0r:
intresting

DevelopmentCats:
You know what I think there may be an unreleased PR for git-clone let me look into it

DevelopmentCats:
Yep it was #1127

35C4n0r:
version bump done.

DevelopmentCats (approved):

Issue #207, opened 2025-07-11
Pre-install VS Code extensions in desktop VS Code module


matifali:
Extend the VS Code desktop module to automatically install specified extensions. Should accept extension list as parameter and configure workspace-specific settings.

yashksaini-coder:
/attempt #207 

My appraoch it to Add an extra extensions parameter that accepts a list of extension IDs. Create a new `run.sh` script that uses `code --install-extension` to install each extension during VM provisioning after VS Code setup. Include basic error handling and ensure proper user context for persistent installations.

yashksaini-coder:
@matifali I have raised PR, for this issue, #218

35C4n0r:
I'm actively working on this, and will raise a PR soon.

bpmct:
Hmm... Why is this challenging?

lsrtech:
/attempt #207

I've submitted PR #926 to address this.

**Approach:** Adds `extensions`, `settings`, and `extensions_dir` variables to the `vscode-desktop` module. On workspace start, a script installs the requested extensions to `~/.vscode-server/extensions/` and merges machine settings into `~/.vscode-server/data/Machine/settings.json`.

This addresses the blocker from the previous attempt (#218): the script auto-detects an existing VS Code CLI and falls back to downloading VS Code Server when none is present, so it works on a fresh workspace. It follows the same per-module pattern already shipped in `vscode-web` and `code-server`, and exposes `extensions_dir` so the same approach can be reused by the Cursor/Windsurf/Kiro wrappers in follow-ups.

Issue #208, opened 2025-07-11
Pre-install JetBrains plugins in Workspaces


matifali:
Add plugin pre-installation capability to [JetBrains IDE module](https://registry.coder.com/modules/coder/jetbrains) (IntelliJ, PyCharm, etc.). Should accept the plugin list as a parameter and configure IDE settings automatically.

Make sure this works with JetBrains Toolbox.

35C4n0r:
@matifali I see no activity on the above PR, can I get assigned to it.

Abiji-2020:
@matifali Just a quick question , that in the JetBrains module, the selected IDEs should  be downloaded on the workspace right ? But currently it just forwards the request to the JetBrains Toolbox right?.  Before installing the complete IDE we cannot create default plugins on it.

NiceBdsmer:
Hi team! I can take this one. I added plugin pre-install support to the JetBrains module and opened a PR here: https://github.com/coder/registry/pull/670.\n\nIt adds a  parameter and a startup script that installs Marketplace plugin IDs once the RemoteDev server is available after the first Toolbox/Gateway connection. Happy to adjust if you want a different install flow.

NiceBdsmer:
Follow-up: PR adds a plugins parameter (list of Marketplace plugin IDs) and installs them once the RemoteDev server is available after first Toolbox/Gateway connection. Happy to tweak behavior if needed.

Rhan2020:
/attempt #208

Opened PR #718 — takes a different approach from previous attempts by downloading plugins directly from the Marketplace API instead of using the IDE CLI. This avoids the race condition with running IDEs that blocked earlier PRs.

kumargope:
/attempt #208 "I'm working on a robust solution using the Marketplace API to pre-install plugins before the IDE starts, ensuring no race conditions."

PR #1094, opened 2026-09-03
feat(vscode-desktop): pre-install remote extensions

Edd88-pixel:
## Why

The VS Code Desktop wrapper could not pre-install remote extensions before the first desktop connection . This advances the extension pre-installation portion of #207 by bootstrapping the official stable remote server and delegating installation to `vscode-desktop-core`.

## Changes

- add an optional validated `extensions` input while preserving the existing empty default
- adopt `vscode-desktop-core` v1.2.0 and pass the remote CLI, config, and extension paths
- download and validate the official stable VS Code Server only when its CLI is absent
- block ordinary workspace login while configured extensions are installed
- document extension configuration and bump the module version to v1.3.0
- cover defaults, validation, rendered configuration, and server reuse

## Validation

- `bun test registry/coder/modules/vscode-desktop/main.test.ts` — 9 passed, 0 failed
- `terraform validate -no-color` — passed
- `terraform fmt -check for the changed Terraform files — passed
- real E2E with Coder v2.33.11

## Description

Adds opt-in remote extension pre-installation to the VS Code Desktop module.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [x] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/vscode-desktop`  
**New version:** `v1.3.0`  

## Testing & Validation

- [x] Tests pass (`bun test`)
- [x] Changes tested locally

## Related Issues

Related to #207

### DIFF
--- a/registry/coder/modules/vscode-desktop/README.md
+++ b/registry/coder/modules/vscode-desktop/README.md
@@ -16,7 +16,7 @@ Uses the [Coder Remote VS Code Extension](https://github.com/coder/vscode-coder)
 module "vscode" {
   count    = data.coder_workspace.me.start_count
   source   = "registry.coder.com/coder/vscode-desktop/coder"
-  version  = "1.2.1"
+  version  = "1.3.0"
   agent_id = coder_agent.main.id
 }
 ```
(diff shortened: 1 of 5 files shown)


DevelopmentCats (approved):
There are a couple things regarding the scorecard that can be addressed in a later PR but otherwiser this looks and works well

matifali:
How do we ensure we are not installing an incompatible extension, since a user may be connecting from an older VS Code release that would download its own server?
AFAIK, VS Code downloads a matching server, not the latest.

It's still a good workaround, though. Let's add this to https://github.com/coder/coder/tree/main/dogfood/coder for dogfooding

matifali:
Also @Edd88-pixel should #208 now be closed?

Edd88-pixel:
@matifali Do you mean #207? #208 is already closed.

For #207, this PR covers the extension pre-installation part, but the settings input/application is still outstanding, so I'd keep it open unless we want to track that separately.

Edd88-pixel:
Good point. The bootstrap server is isolated, so an older VS Code client can still install its own matching server. The part we don't currently guarantee is extension compatibility, since extensions are initially resolved using the stable bootstrap server.

Dogfooding this makes sense. We can follow up on that and test the behavior with older VS Code versions.

matifali:
> @matifali Do you mean #207? #208 is already closed.
> 
> For #207, this PR covers the extension pre-installation part, but the settings input/application is still outstanding, so I'd keep it open unless we want to track that separately.

Yes I mean #207

PR #1108, opened 2026-09-09
feat(claude-code): add authentication_config to keep gateway auth out of the shell env

bpmct:
With `enable_ai_gateway = true`, the module delivers `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` as `coder_env` resources. That works, but it means the workspace owner's session token shows up in plain `env` output in every shell. For deployments (and demos) whose whole point is "the workspace holds no AI credentials", that undercuts the story even though the gateway is doing the right thing.

This adds an `authentication_config` input:

- `"environment"` (default): current behavior, unchanged.
- `"managed_settings"`: writes those two values into the `env` block of `/etc/claude-code/managed-settings.d/10-coder.json` instead. Claude Code reads env from settings files, and managed settings are the highest-precedence location, so the CLI behaves identically while `env` in the shell stays clean.

```tf
module "claude-code" {
  source                = "registry.coder.com/coder/claude-code/coder"
  version               = "5.5.0"
  agent_id              = coder_agent.main.id
  enable_ai_gateway     = true
  authentication_config = "managed_settings"
}
```

Scope is deliberately narrow: only `ANTHROPIC_BASE_URL` and `ANTHROPIC_AUTH_TOKEN` move. `ANTHROPIC_MODEL` isn't authentication and stays a `coder_env`. `anthropic_api_key`, `claude_code_oauth_token`, Bedrock/Vertex flags, and telemetry are untouched. A user-supplied `managed_settings.env` block is preserved; the gateway keys are merged in on top.

The README is honest that this isn't a secrecy boundary: the token still lands in a root-owned `0644` file. It removes the value from the process environment, nothing more.

One implementation note for reviewers: `managed_settings_effective` uses a `jsonencode`/`jsondecode` round-trip around the ternary. A plain `cond ? merge(...) : var.managed_settings` fails `terraform test` with `Inconsistent conditional result types` whenever a concrete `managed_settings` value lacks an `env` key, because both branches of an object-valued conditional must share a structural type. Encoding both branches to strings sidesteps that; the comment in `main.tf` cites the exact error. Happy to take a cleaner approach if someone knows one.

Tests: 4 new `plan` runs (default keeps the `coder_env`s; `managed_settings` yields zero of them and the right two keys; user `env` keys preserved; invalid value rejected). `terraform test` is 31/31, `terraform validate` and `bun run fmt` clean. Version bumped minor to 5.5.0.

Motivation came from wiring this up by hand in [coder/workshops](https://github.com/coder/workshops) for the Coder Way track; this makes `enable_ai_gateway = true` sufficient on its own.

🤖 Generated with Coder Agents on behalf of @bpmct


DevelopmentCats (approved):
tis good now

PR #1112, opened 2026-09-11
feat(agent-relay-cursor): add the Agent Relay Cursor module

Emyrk:
## Description

Adds `coder/agent-relay-cursor`, the template module for [Agent Relay](https://coder.com/docs/ai-coder/agent-relay) Cursor pools. It declares the `agent_relay_*` parameter contract the relay stamps on every build and verifies at startup, exports the worker id as `CURSOR_AGENT_WORKER_ID` and the per-user worker token Agent Relay mints (coder/agent-relay#60) as `AGENT_RELAY_CURSOR_TOKEN`, passed to the CLI as `--auth-token`, and runs `agent worker ... start` detached, publishing the worker lifecycle through an `agent_relay_status` agent metadata item that the template must declare on its `coder_agent`.

Moved from [coder/agent-relay](https://github.com/coder/agent-relay/tree/main/module/cursor), where it was developed and tested against the daemon; that copy is retired in favor of this one.

- Every relay parameter renders disabled with a "Set by Agent Relay on dispatch" placeholder; the credential is masked.
- `install_cli` defaults to `true` and only downloads when the CLI is not already on PATH.
- `main.tftest.hcl` covers the parameter contract and the rendered scripts; `main.test.ts` runs the rendered script in a container with a stub `agent` binary and observes the idle, failed, working, and done states (#1113).

- Pool name and idle timeout reach the worker through `coder_env` (`AGENT_RELAY_CURSOR_POOL_NAME`, `AGENT_RELAY_CURSOR_IDLE_RELEASE_TIMEOUT`) as quoted variables, never interpolated into the script, so parameter values cannot run as shell. `main.test.ts` dispatches a pool name containing shell metacharacters and asserts it arrives as one argument.

- Scripts run through `coder-utils` as an install step (no-op when the CLI is present) and a start step. Scripts, step logs, runner state, and runner log all live under `$HOME/.coder-modules/coder/<module>` for debugging.

## Type of Change

- [x] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/agent-relay-cursor`
**New version:** `v0.1.0`

## Testing & Validation

- [x] Tests pass (`terraform test` 7 runs, `bun test` 9 tests)
- [x] Changes tested locally (`readmevalidation`, `terraform validate`)

## Related Issues

None

---

Closes RELAY-18

### DIFF
--- a/registry/coder/modules/agent-relay-cursor/README.md
+++ b/registry/coder/modules/agent-relay-cursor/README.md
@@ -1,0 +1,169 @@
+---
+display_name: Agent Relay Cursor
+description: Serves Cursor cloud agent requests in Coder workspaces that Agent Relay dispatches.
+icon: ../../../../.icons/cursor.svg
+verified: true
+tags: [agent, cursor, agent-relay]
+---
+
+# Agent Relay Cursor
+
+Makes a Coder template a target for [Agent Relay](https://coder.com/docs/ai-coder/agent-relay)
+Cursor pools: include the module, wire it to your `coder_agent`, and Agent
+Relay can dispatch [Cursor cloud agent](https://coder.com/docs/ai-coder/agent-relay/cursor)
+requests to workspaces built from the template.
+
+```tf
+module "cursor_worker" {
+  source   = "registry.coder.com/coder/agent-relay-cursor/coder"
+  version  = "1.0.0"
+  agent_id = coder_agent.main.id
+
+  # Default. Installs the CLI only when the image does not already
+  # carry it; baking it into the image stays the fastest path.
+  install_cli = true
+}
+
+resource "coder_agent" "main" {
+  # ...
+  metadata {
+    key          = "agent_relay_status"
+    display_name = "Worker"
+    script       = module.cursor_worker.status_metadata_script
+    interval     = 10
+    timeout      = 5
+  }
+}
+```
+
+The `agent_relay_status` metadata block is required and must live on the
+agent: the coder provider has no standalone agent-metadata resource, so
+a module cannot declare it. Referencing the module's
+`status_metadata_script` output does not create a dependency cycle; the
+output does not depend on `agent_id`.
+
+Requirements for the template and image:
+
+- The Cursor CLI (`agent`) must be available in the workspace.
+  `install_cli` defaults to `true` and downloads the latest CLI at
+  workspace start, but only when the binary is not already on PATH: a
+  CLI in the image is used as is, never upgraded, and costs nothing.
+  When the download does run it needs outbound access to cursor.com and
+  spends part of the claim-to-ready window. `cli_binary` overrides the
+  path. With no CLI and no successful install, the worker reports
+  `failed runner-agent-missing`, which
+  Agent Relay reaps and records.
+- For repo-scoped pools the template must clone the repository named
+  by `agent_relay_cursor_repo_url` and provide SCM credentials before the worker
+  starts, e.g. via a startup script ordered before this module's
+  script. Repo-less pools should provide a `.cursor/rules` file in the
+  worker's working directory instead.
+- With `computer_use = true` the image must also carry the
+  computer-use packages the Cursor CLI expects.
+- Builds should complete within the pool's `dispatch_deadline`
+  (default 10m): pre-pulled images, no persistent volumes. The
+  deadline cannot exceed 15m: Agent Relay's reaper deletes
+  workspaces stuck pending or starting after that grace, and it does
+  so even when `dispatch_deadline` is `"0"` (no dispatcher deadline).
+
+## Credential exposure
+
+The pool's service-account API key reaches every dispatched workspace:
+it is stamped as the ephemeral `agent_relay_credential` build parameter
+and exported as the `CURSOR_API_KEY` environment variable, where the
+workspace owner can read it for the lifetime of the workspace. Treat
+the key accordingly:
+
+- Use a dedicated service-account key per pool, scoped to the
+  repository the pool serves, distinct from any broader Cursor
+  key your team uses.
+- Rotating the key in Agent Relay's config only affects new builds;
+  running workspaces keep the value they were built with until reaped.
+
+Sub-token worker authentication (short-lived per-worker credentials
+instead of the pool key) is the tracked hardening follow-up.
+
+The module declares the full parameter contract Agent Relay stamps on
+every build; Agent Relay verifies the contract against the template's
+active version at startup (dynamic parameters evaluate endpoint) and
+refuses to serve a pool whose template does not satisfy it:
+
+| parameter                                 | kind             | value                                                                                                                      |
+| ----------------------------------------- | ---------------- | -------------------------------------------------------------------------------------------------------------------------- |
+| `agent_relay_session_id`                  | persistent state | Cursor request served by this workspace (dedupe/reconciliation key)                                                        |
+| `agent_relay_delivery_id`                 | persistent state | worker id the request was claimed with; exported as `CURSOR_AGENT_WORKER_ID`                                               |
+| `agent_relay_pool`                        | persistent state | Agent Relay pool that dispatched this build                                                                                |
+| `agent_relay_cursor_pool_name`            | persistent state | Cursor-side pool the worker registers under                                                                                |
+| `agent_relay_cursor_idle_release_timeout` | persistent state | seconds the worker idles after the agent finishes a turn before releasing and exiting; keep it at or above 300 (see below) |
+| `agent_relay_cursor_repo_url`             | persistent state | repository the request targets; empty for repo-less pools, but always declared so the contract stays static                |
+| `agent_relay_credential`                  | ephemeral        | service-account API key the worker authenticates with; exported as `CURSOR_API_KEY`                                        |
+
+The persistent parameters matter because coderd only stores values the
+template declares; Agent Relay's dedupe and reconciliation depend on
+querying them with `param:` search filters.
+
+Every parameter above renders disabled, with a placeholder saying Agent Relay
+sets it. That is cosmetic, not a guardrail: it keeps a person creating a
+workspace by hand from filling in machine-set fields, and does not stop the
+CLI or the API from sending values. The credential is masked for the same
+reason.
+
+## Worker lifecycle reporting
+
+Agent Relay deletes a workspace once its worker exits, and records the
+failure when the worker died instead; Cursor offers no nack, so
+observability is the only failure signal. Neither Coder nor Agent Relay
+knows what a Cursor session is, so the module publishes the worker's
+lifecycle and Agent Relay reads it.
+
+The worker script starts
+
+```sh
+CURSOR_API_KEY= < token > CURSOR_AGENT_WORKER_ID= < worker id > \
+agent worker --pool < agent_relay_cursor_pool_name > \
+  --idle-release-timeout < idle_release_timeout > start
+```
+
+detached and exits, so the agent reaches the `ready` lifecycle state
+instead of sitting in `starting` for the life of the session. The
+worker exits `0` on its own once the idle-release timer fires after a
+session ends; that clean exit is what tells Agent Relay to reap. Mind the
+timer semantics: it starts when the agent finishes a turn, not when
+the user closes the chat, so `idle_release_timeout` must cover the
+user's reading and typing time. Once released, a follow-up to that
+chat errors on Cursor's side until Cursor issues a new pending request
+(which Agent Relay then serves with a fresh workspace). Keep the timeout
+at or above 300 seconds (Agent Relay warns below that; the default is
+600). A
+detached supervisor owns the worker process and writes single-line
+state to `state_file` (`/tmp/agent-relay/worker-state` by default).
+The `agent_relay_status` metadata item reports one of:
+
+| value             | meaning                                                                                                 |
+| ----------------- | ------------------------------------------------------------------------------------------------------- |
+| `pending`         | supervisor has not recorded state yet (agent just started)                                              |
+| `idle`            | no worker token: workspace was created manually                                                         |
+| `working`         | worker alive, no session attached yet                                                                   |
+| `serving`         | worker alive and a session has attached                                                                 |
+| `orphaned`        | state says working but the process is gone, and the supervisor died too                                 |
+| `done <code>`     | worker exited with that status (`0` is the normal idle release; `137` means it was killed)              |
+| `failed <reason>` | the worker could not be started at all, e.g. `runner-agent-missing` when the binary is not in the image |
+
+`working` versus `serving` comes from matching `serving_log_pattern`
+against the worker log. At default verbosity the Cursor CLI log
+carries no session-attach line at all, so the pattern can only match
+with verbose worker logs and typically never fires. That is fine:
+Agent Relay's status page overlays Cursor's authoritative in-use worker
+state (from the fleet worker listing) on top of this metadata, so a
+workspace whose worker is in use renders as serving either way, and
+reaping decisions rest on the process lifecycle. The metadata-level
+distinction is therefore best-effort and purely cosmetic; tune the
+pattern only if you run verbose logs and want the distinction inside
+the workspace metadata too.
+
+The script's timings row is not the reaping signal. coderd records it
+when the script ends, which is now seconds after the agent starts, long
+before the session finishes.
+
+The `agent_relay_status` key is part of the contract; renaming it breaks
+reaping.
--- a/registry/coder/modules/agent-relay-cursor/README.md
+++ b/registry/coder/modules/agent-relay-cursor/README.md
@@ -18,6 +18,10 @@ module "cursor_worker" {
   source   = "registry.coder.com/coder/agent-relay-cursor/coder"
   version  = "1.0.0"
   agent_id = coder_agent.main.id
+
+  # Downloads the Cursor CLI at start when it is not in the image. Bake
+  # the CLI into the image and set this to false for faster workspaces.
+  install_cli = true
 }
 
 resource "coder_agent" "main" {
--- a/registry/coder/modules/agent-relay-cursor/README.md
+++ b/registry/coder/modules/agent-relay-cursor/README.md
@@ -16,7 +16,7 @@ parameters the relay stamps on each build and runs the Cursor CLI worker.
 ```tf
 module "cursor_worker" {
   source   = "registry.coder.com/coder/agent-relay-cursor/coder"
-  version  = "1.0.0"
+  version  = "0.1.0"
   agent_id = coder_agent.main.id
 
   # Downloads the Cursor CLI at start when it is not in the image. Bake
--- a/registry/coder/modules/agent-relay-cursor/README.md
+++ b/registry/coder/modules/agent-relay-cursor/README.md
@@ -9,19 +9,15 @@ tags: [agent, cursor, agent-relay]
 # Agent Relay Cursor
 
 Makes a Coder template a target for [Agent Relay](https://coder.com/docs/ai-coder/agent-relay)
-Cursor pools: include the module, wire it to your `coder_agent`, and Agent
-Relay can dispatch [Cursor cloud agent](https://coder.com/docs/ai-coder/agent-relay/cursor)
-requests to workspaces built from the template.
+Cursor pools. Agent Relay dispatches [Cursor cloud agent](https://coder.com/docs/ai-coder/agent-relay/cursor)
+requests to workspaces built from the template; the module declares the
+parameters the relay stamps on each build and runs the Cursor CLI worker.
 
 ```tf
 module "cursor_worker" {
   source   = "registry.coder.com/coder/agent-relay-cursor/coder"
   version  = "1.0.0"
   agent_id = coder_agent.main.id
-
-  # Default. Installs the CLI only when the image does not already
-  # carry it; baking it into the image stays the fastest path.
-  install_cli = true
 }
 
 resource "coder_agent" "main" {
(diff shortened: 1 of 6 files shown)


registry/coder/modules/agent-relay-cursor/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```tf
+module "cursor_worker" {
+  source   = "registry.coder.com/coder/agent-relay-cursor/coder"
+  version  = "0.1.0"
+  agent_id = coder_agent.main.id
+
+  # Downloads the Cursor CLI at start when it is not in the image. Bake
+  # the CLI into the image and set this to false for faster workspaces.
+  install_cli = true
```
matifali: Can we be a bit intelligent and auto-skip downloading if we find it on the path?

registry/coder/modules/agent-relay-cursor/run.sh.tftpl
```diff
@@ -0,0 +1,114 @@
...
+write_state() {
+	printf '%s\n' "\$1" >"$state_file.tmp"
+	mv "$state_file.tmp" "$state_file"
+}
+
+%{ if install_cli ~}
+export PATH="\$HOME/.local/bin:\$PATH"
+%{ endif ~}
+${cli_binary} worker \\
+	--pool "${pool_name}" \\
```
Edd88-pixel, line 84: This parameter is interpolated directly into the generated script. A value containing shell metacharacters can break out of the quoted argument and execute a command with access to `CURSOR_API_KEY`. For example, `safe"; command; #` terminates the argument and executes `command` in the worker environment, which includes the Cursor API key.

Could we encode the value before passing it to the script, then decode it into a properly quoted variable before invoking the worker?

registry/coder/modules/agent-relay-cursor/run.sh.tftpl
```diff
@@ -0,0 +1,114 @@
...
+	--idle-release-timeout "${idle_release_timeout}" \\
```
Edd88-pixel, line 85: The same issue applies to `idle_release_timeout` on this line.

Edd88-pixel (changes requested):
I suggest preventing these values from being interpreted as shell code.

registry/coder/modules/agent-relay-cursor/run.sh.tftpl
Emyrk: Fixed in 95076fc. Both values now travel through `coder_env` and are read as quoted variables at run time, like the token; nothing from a parameter is rendered into the script. Added a test that dispatches `safe\"; touch /tmp/injected; #` as the pool name and asserts it arrives as one argument.

<sub>Coder Agents on behalf of @Emyrk.</sub>

Edd88-pixel: Thanks, this addresses the issue.

registry/coder/modules/agent-relay-cursor/start.sh.tftpl
```diff
@@ -0,0 +1,110 @@
...
+for _ in 1 2 3 4 5 6 7 8 9 10; do
+	[ -f "$state_file" ] && break
+	sleep 1
+done
+
+if [ -f "$state_file" ]; then
+	echo "Worker state: $(cat "$state_file")"
+else
+	echo "Worker did not report state within 10s; check $log_file" >&2
+fi
```
matifali, line 145: **Stale state is never cleared on the credential path, so this loop can report a previous run's terminal state.**

The no-credential branch above deliberately preserves an existing state file (good — the `done 3` test covers it), but the credential branch never resets it. `[ -f "$state_file" ]` is then satisfied by the *stale* file, so this loop breaks on the first iteration without ever waiting for the supervisor.

I rendered the template with the defaults and ran it with a pre-existing state file:

```
$ echo "done 3" > ~/.coder-modules/coder/agent-relay-cursor/worker-state
$ bash status.sh                       # what the relay's metadata poll sees
done 3
$ AGENT_RELAY_CURSOR_TOKEN=tok ... bash start.sh
Starting Cursor worker (detached)...
Worker state: done 3                   # <-- a worker was just launched
$ bash status.sh                       # immediately after start.sh returned
done 3
$ sleep 1 && bash status.sh
working
```

So `agent_relay_status` publishes a terminal `done 3` for the whole window between agent start and the supervisor's first write. That window is not always sub-second: with `install_cli = true` and no CLI in the image, the `cursor.com/install` download runs before this script, and the metadata interval in the README example is 10s. If the reaper acts on `done <code>`, it can delete a workspace whose worker is starting up.

The trigger doesn't need a persistent volume or workspace reuse — an **agent restart within the same build** is enough: `coder-utils` sets `run_on_start = true` on its scripts, and the credential arrives via `coder_env`, which is build-scoped, so the restarted agent re-runs this script with `AGENT_RELAY_CURSOR_TOKEN` still set and the old `worker-state` still on disk.

A related consequence of the same root cause: a stale `working <pid>` also survives that restart, and `status.sh` only tests `kill -0 "$detail"`. I confirmed it reports `working` for an unrelated live PID (wrote `working <pid-of-a-sleep>` and got `working`), so PID reuse can make the relay believe a worker is alive indefinitely.

Fix looks like one line in the credential branch, before the supervisor is written/launched:

```bash
# A restart re-runs this script with the build's credential still set; the
# previous run's state must not be read as this run's.
write_state "pending"
```

That makes this loop meaningful again and keeps `pending` — already in the documented vocabulary — as the pre-supervisor state.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/start.sh.tftpl
```diff
@@ -0,0 +1,110 @@
...
+# The supervisor outlives this script: it owns the worker process and
+# is the only writer of terminal state. Written to disk rather than
+# inlined so setsid gets a clean argv. The token, pool name, and idle
+# timeout are read from the supervisor's environment at run time,
+# escaped here so they are never rendered into the file. Parameter
+# values are user input; interpolating them into the file would let
+# shell metacharacters run as code. The worker exits 0 when its
+# idle-release timer fires after a session ends; that clean exit is
+# what tells Agent Relay to reap the workspace.
+supervisor="$state_dir/supervise.sh"
```
matifali: **`supervise.sh` is placed relative to `state_file`, so it escapes the module directory when that variable is overridden.**

`state_dir` comes from `dirname "$state_file"` (line 17), so overriding `state_file` moves the supervisor with it. With `state_file = "/var/lib/relay/state"` — the exact override `main.test.ts` exercises in *"uses cli_binary and state_file overrides"* — the supervisor is written to `/var/lib/relay/supervise.sh`. That contradicts the README's own layout table, which documents `supervise.sh` under `$HOME/.coder-modules/coder/agent-relay-cursor`.

Even with the defaults it lands at the module *root* rather than in `scripts/`, where `AGENTS.md` puts runtime-materialized scripts:

> | `scripts/` | Scripts materialized at runtime (if any) | `$HOME/.coder-modules/coder/claude-code/scripts/install.sh` |

Suggest decoupling the two: pass `module_directory` into the template and write the supervisor to `$module_directory/scripts/supervise.sh`. `state_file` then only controls state, which is what its description promises, and the README table stays accurate for any override.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/main.tf
```diff
@@ -0,0 +1,269 @@
...
+    install_cli  = var.install_cli
+    computer_use = var.computer_use
+    state_file   = var.state_file
+    log_file     = var.log_file
+  })
+}
+
+module "coder_utils" {
+  source  = "registry.coder.com/coder/coder-utils/coder"
+  version = "0.0.1"
```
matifali, line 251: **Missing the `scripts` pass-through output.**

`AGENTS.md` requires it for every `coder-utils` consumer:

> Always expose the `scripts` output as a pass-through so upstream modules can serialize their own `coder_script` resources behind this module's install pipeline

This module is the only one of the 15 `coder-utils` consumers in the registry without it (`grep -c 'output "scripts"'` is 1 for `claude-code`, `agent-firewall`, `supabase`, `pool`, `codex`, `mcp-servers`, and the rest — 0 here). Worth adding for consistency, especially since a Cursor template will plausibly want to chain a repo-clone step behind the worker install:

```tf
output "scripts" {
}
```

`main.tftest.hcl:104` already asserts on `module.coder_utils.scripts`, so the assertion could move to `output.scripts` and cover the contract callers actually consume.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/main.tf
```diff
@@ -0,0 +1,269 @@
...
+
+variable "state_file" {
+  type        = string
+  default     = "$HOME/.coder-modules/coder/agent-relay-cursor/worker-state"
+  description = "Path the worker supervisor writes its lifecycle state to, read by the agent_relay_status agent metadata item."
+}
+
+variable "log_file" {
+  type        = string
+  default     = "$HOME/.coder-modules/coder/agent-relay-cursor/logs/worker.log"
```
matifali, line 79: Nit: an empty `serving_log_pattern` inverts the documented fallback. `grep -qF -- ""` matches any non-empty file, so every poll reports `serving`:

```
with default pattern 'in use': working
with empty pattern:            serving
```

The description promises the opposite — *"A pattern that never matches degrades to working and affects nothing else"* — so an empty value is the one input that breaks the stated safety property. A validation block would keep the guarantee honest:

```tf
  validation {
    condition     = length(var.serving_log_pattern) > 0
    error_message = "serving_log_pattern must not be empty; an empty pattern matches every log line."
  }
```

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/README.md
```diff
@@ -0,0 +1,117 @@
...
+- Builds must finish inside the pool's `dispatch_deadline` (default 10m, max
+  15m): pre-pulled images, no persistent volumes.
+
+## Worker credential
+
+The pool's service-account API key never leaves Agent Relay. At dispatch the
+relay exchanges it for a Cursor sub-token scoped to the requesting user and
+stamps that as the ephemeral `agent_relay_credential` parameter. The module
+exports it as `AGENT_RELAY_CURSOR_TOKEN` and starts the worker with
+`--auth-token`; it is never written to disk. The token acts only as that user,
```
matifali: Doc nit: *"it is never written to disk"* is accurate and well enforced (the supervisor file references `$AGENT_RELAY_CURSOR_TOKEN` rather than the value, and `main.test.ts` asserts the literal token is absent from `supervise.sh` — nice), but `--auth-token` puts the token in the worker's `/proc/<pid>/cmdline`, which is readable by any process running as that user in the workspace.

The `main.tf` comment explains the CLI offers no env var alternative, so this is a Cursor CLI constraint rather than a module bug. Still, since this section makes an explicit hygiene claim, one clause would set the boundary correctly for operators — e.g. "never written to disk, though it is visible in the worker's process arguments inside the workspace."

---
_Generated by [Claude Code](https://claude.ai/code)_

matifali (commented):
Reviewed the module end to end. Since neither `terraform` nor `bun` is on this runner, I rendered `start.sh.tftpl` / `status.sh.tftpl` with a faithful `templatefile()` stand-in (verified against the assertions in `main.tftest.hcl` and `main.test.ts`) and executed the rendered scripts to confirm behavior rather than reasoning from the diff. Every finding below is reproduced, not inferred.

**Overall this is high-quality work.** The injection handling is genuinely solid — routing parameter values through `coder_env` and reading them as quoted variables, base64-encoding the free-form pattern and matching with `grep -F`, and keeping the token out of the supervisor file are the right calls, and each is backed by a test that actually attempts the attack. I re-ran the `serving_log_pattern` case and confirmed `/tmp/PWNED` is never created. The test suite is better than most modules in this registry.

One substantive issue and four smaller ones:

1. **`start.sh.tftpl`** — stale worker state is never cleared on the credential path. The readiness loop short-circuits on a previous run's state file and `agent_relay_status` publishes a terminal `done <code>` while a fresh worker is starting. The trigger needs no persistent volume: an agent restart within the same build re-runs the start script with the build's `coder_env` credential still set and the old `worker-state` on disk. Reproduced — details inline. This is the one I'd want resolved before merge.
2. **`start.sh.tftpl`** — `supervise.sh` is placed via `dirname "$state_file"`, so the `state_file` override your own test uses writes it to `/var/lib/relay/supervise.sh`, outside the module directory and contrary to the README table. Also sits at the module root instead of `scripts/` per `AGENTS.md`.
3. **`main.tf`** — missing the `scripts` pass-through output that `AGENTS.md` mandates; this is the only one of 15 `coder-utils` consumers without it.
4. **`main.tf`** — an empty `serving_log_pattern` makes every poll report `serving`, the inverse of what the variable description guarantees.
5. **`README.md`** — the "never written to disk" claim is true, but the token is visible in the worker's process arguments.

Two notes on things I checked and am *not* flagging: `verified: true` matches the coder namespace convention (49 of 51 modules), and `0.1.0` is a fine initial version (`agent-firewall` is at `0.0.3`, `coder-utils` at `0.0.1`). All five CI checks are green.

Finally — my earlier open thread asking whether the install could auto-skip when the CLI is already on PATH is **fully addressed**: `install.sh.tftpl` guards on `command -v`, `start.sh.tftpl` adds `~/.local/bin` unconditionally so a bring-your-own CLI is found, and there are tests for both. I'll leave that thread for you to resolve.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/start.sh.tftpl
matifali: **Correcting myself on two points above.** I re-measured, and I got the severity distribution backwards. The root cause is real, but the durable failure is in the *other* branch.

**1. I overstated the credential-path window.** I claimed it "is not always sub-second" because the `cursor.com/install` download runs first. That reasoning is wrong for the default configuration: `state_file` defaults under `$HOME` and the installer writes to `$HOME/.local/bin`, so the stale state file and the CLI share a filesystem root — whenever stale state survives a restart, the CLI has survived too and the download short-circuits. Measured `start.sh`'s own contribution across three runs:

```
run 1: stale state observable for 12ms
run 2: stale state observable for 15ms
run 3: stale state observable for  8ms
```

So the supervisor overwrites within ~10ms. A 10s metadata poll landing in that window is unlikely, and my "can delete a workspace whose worker is starting up" framing was too strong. It only stretches if someone points `state_file` at a persistent path while `$HOME` is ephemeral — an override case, not the default.

**2. I attached the PID-reuse consequence to the wrong branch.** I said stale `working <pid>` could make the relay believe a worker is alive "indefinitely." Not on this path — the supervisor overwrites it in ~10ms. It *is* indefinite on the **no-credential** path, i.e. the branch at lines 26-36 that deliberately preserves state. A user-initiated stop/start leaves the ephemeral credential empty, so that branch runs, keeps the stale line, and never launches a worker:

```
# stale 'working <DEAD pid>', no credential
  preserved as: working 999999    -> status: orphaned          (defensible, documented)

# stale 'working <REUSED live pid>', no credential
  preserved as: working 15293     -> status: working
  PID 15293 cmdline: sleep 120                                 (not a worker at all)
```

Nothing overwrites that file, so `agent_relay_status` reports `working` indefinitely while no worker exists. `status.sh` only tests `kill -0 "$detail"`, which answers "is *some* process alive," not "is *the worker* alive." After a restart the container's PIDs count up from low numbers, so landing on a previously-recorded worker PID is plausible rather than exotic.

**What this means for the fix.** My suggested `write_state "pending"` in the credential branch is still worth having — it makes this readiness loop meaningful instead of short-circuiting on a stale file — but it does **not** address the case above. That one wants a PID identity check in `status.sh`, e.g. confirming `/proc/<pid>/cmdline` actually belongs to the CLI before reporting `working`, and otherwise falling through to `orphaned`. I verified the two are distinguishable — a real worker's argv is `... agent worker --pool p ... start` while the decoy's is `sleep 120`.

Net: treat the credential-path issue as a minor correctness/cosmetic fix, and the no-credential stale `working` as the one worth a decision before merge. Apologies for the misdirection in my original comment — the other four findings I raised are unchanged and re-verified.

---
_Generated by [Claude Code](https://claude.ai/code)_

matifali (commented):
## Review summary (supersedes the severity ordering in my earlier review)

I re-validated every finding after posting, and one of them changed materially — my first review called the credential-path stale state "the one I'd want resolved before merge," and that was wrong. Consolidating here so the accurate picture is at top level rather than inside a thread reply.

**Method:** neither `terraform` nor `bun` is available on my runner, so I rendered `start.sh.tftpl` / `status.sh.tftpl` with a `templatefile()` stand-in (validated against the assertions in `main.tftest.hcl` and `main.test.ts`) and executed the rendered scripts. Nothing below is inferred from reading the diff alone.

### Worth a decision before merge

**Stale `working <pid>` on the no-credential path reports a live worker when none exists** — [thread](https://github.com/coder/registry/pull/1112#discussion_r4038884765).

The branch at `start.sh.tftpl:26-36` deliberately preserves an existing state file, which is correct for the `done 3` case your test covers. But a user-initiated stop/start leaves the ephemeral credential empty, so that branch also runs after a session where the worker was still marked `working` — it keeps the stale line and never launches a worker. `status.sh:38` then only tests `kill -0 "$detail"`, which answers "is *some* process alive," not "is *the worker* alive":

```
stale 'working <DEAD pid>', no credential
  preserved as: working 999999  -> status: orphaned      (defensible, documented)

stale 'working <REUSED live pid>', no credential
  preserved as: working 15293   -> status: working
  PID 15293 cmdline: sleep 120                           (not a worker at all)
```

Nothing overwrites that file on this path, so `agent_relay_status` can report `working` indefinitely with no worker running. After a restart the container's PIDs count up from low numbers, so colliding with a previously-recorded worker PID is plausible rather than exotic. A PID identity check in `status.sh` — confirming `/proc/<pid>/cmdline` belongs to the CLI before reporting `working`, else falling through to `orphaned` — closes it. I verified the two are distinguishable (a real worker's argv is `agent worker --pool p ... start`).

### Minor correctness

**The readiness loop short-circuits on a stale state file** (same thread). The credential branch never resets state, so `[ -f "$state_file" ]` at `start.sh.tftpl:101` is satisfied by the previous run's file and the loop breaks on iteration one — printing `Worker state: done 3` for a run that just launched a fresh worker. The trigger is an agent restart within the same build (`coder-utils` sets `run_on_start = true`; the credential arrives via build-scoped `coder_env`).

To correct my own earlier claim: this window is **8-15ms**, not the multi-second window I first described. I had argued the `cursor.com/install` download extends it, but `state_file` and the installer's `~/.local/bin` share `$HOME`, so whenever stale state survives, the CLI survived too and the download short-circuits. A `write_state "pending"` before launching the supervisor is still worth it for the misreported log line, but this is cosmetic, not a reaping hazard.

### Conventions

- **`supervise.sh` is placed via `dirname "$state_file"`** — [thread](https://github.com/coder/registry/pull/1112#discussion_r4038886123). The `state_file = "/var/lib/relay/state"` override your own `main.test.ts` exercises writes it to `/var/lib/relay/supervise.sh`, outside the module directory and contrary to the README layout table. Even at defaults it lands at the module root rather than `scripts/` per `AGENTS.md`.
- **Missing the `scripts` pass-through output** — [thread](https://github.com/coder/registry/pull/1112#discussion_r4038887378). Exact count: 15 `coder-utils` consumers in the registry, 14 expose it, this is the only one that doesn't.

### Nits

- **Empty `serving_log_pattern` inverts the documented fallback** — [thread](https://github.com/coder/registry/pull/1112#discussion_r4038888381). `grep -qF -- ""` matches any non-empty log, so every poll reports `serving`, the opposite of what the variable description guarantees.
- **Token is in the worker's argv** — [thread](https://github.com/coder/registry/pull/1112#discussion_r4038889741). The README's "never written to disk" is accurate and well enforced; I confirmed by dumping the stub's own `/proc/$$/cmdline` that `--auth-token <value>` is nonetheless readable in-workspace. A Cursor CLI constraint, not a module bug — just worth a clause so the boundary is clear.

### Things I checked and am deliberately not flagging

- `verified: true` matches the coder namespace convention (49 of 51 modules).
- `0.1.0` is a fine initial version (`agent-firewall` is `0.0.3`, `coder-utils` `0.0.1`).
- `description`-before-`type` variable ordering in `AGENTS.md` — every neighboring coder module uses `type` first, so flagging it here would be noise.
- All five CI checks green.

### Credit where due

The injection handling is the strongest part of this PR and I tried to break it without success: routing parameter values through `coder_env` and reading them as quoted variables, base64-encoding the free-form pattern and matching with `grep -F`, and keeping the literal token out of `supervise.sh` are each the right call and each backed by a test that actually attempts the attack. I re-ran the `serving_log_pattern` case and confirmed `/tmp/PWNED` is never created. The test suite is better than most modules in this registry — the two issues above are precisely the kind that live in the gap between "the script is safe" and "the state machine survives a restart," which is a harder target.

Also confirming my earlier question about auto-skipping the download when the CLI is already on PATH is fully addressed: `install.sh.tftpl` guards on `command -v`, `start.sh.tftpl` adds `~/.local/bin` unconditionally so a bring-your-own CLI is found, and both have tests.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-cursor/start.sh.tftpl
Emyrk: Agreed on the diagnosis. Fixed in f1640ce, going one step further than the `pending` reset: in the restart scenario you describe the detached supervisor and worker are normally still alive (only a container restart takes them down, and that also clears the state file), so a blind relaunch would start a second worker against a single-use credential.

The credential branch now reads existing state first:
- `working <pid>` with a live worker: log and exit, no second launch.
- `working <pid>` dead, or `done`/`failed`: leave the state for the relay, exit.
- otherwise: `write_state \"pending\"` and launch, as you suggested.

On the pid-reuse point, both scripts now share a `worker_alive` check that requires `/proc/<pid>/cmdline` to name the CLI in addition to `kill -0`, so a reused pid reports `orphaned`.

Container tests cover each branch: terminal state not restarted, live worker not duplicated (exactly one process), reused pid reads `orphaned`, stale `pending` relaunches.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-cursor/main.tf
Emyrk: Already on the branch as of d59cb48 (landed with the same comment on #1114): `output \"scripts\"` re-exports `module.coder_utils.scripts`, and the tftest asserts `output.scripts == module.coder_utils.scripts` alongside the ordering check.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-cursor/start.sh.tftpl
Emyrk: Fixed in 36f338d: `module_directory` is passed into the start template and the supervisor is written to `$module_directory/scripts/supervise.sh`; `state_file` now controls only the state. README table updated, and the override test asserts the supervisor stays under the module directory while the state moves.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-cursor/main.tf
Emyrk: Added in 8149568, with a tftest using `expect_failures` on the empty value.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-cursor/README.md
Emyrk: Agreed; added the clause in 71bb607.

<sub>Coder Agents on behalf of @Emyrk.</sub>

matifali (approved):

PR #1114, opened 2026-09-11
feat(agent-relay-claude-code): add the Agent Relay Claude Code module

Emyrk:
## Description

Adds `coder/agent-relay-claude-code`, the template module for [Agent Relay](https://coder.com/docs/ai-coder/agent-relay) Claude Code pools. It declares the `agent_relay_*` parameter contract the relay stamps on every build and verifies at startup, exports the Claude Code CLI's `SELF_HOSTED_RUNNER_POOL_SECRET` / `SELF_HOSTED_RUNNER_LOCK_TO_ACCOUNT`, and runs `claude self-hosted-runner` detached through a wrapper that forces `--permission-mode bypassPermissions`, publishing the runner lifecycle through an `agent_relay_status` agent metadata item that the template must declare on its `coder_agent`.

Moved from [coder/agent-relay](https://github.com/coder/agent-relay/tree/main/module/claude-code), where it was developed and tested against the daemon. Same layout and README structure as `agent-relay-cursor` (#1112).

- Every relay parameter renders disabled with a "Set by Agent Relay on dispatch" placeholder; the credential is masked.
- `install_cli` defaults to `true` and only downloads when the CLI is not already on PATH.
- `main.tftest.hcl` covers the parameter contract and rendered scripts.
- `main.test.ts` runs the rendered script in a container with a stub `claude` binary: `idle` without a credential, `failed runner-agent-missing` without a CLI, install skipped when a CLI is on PATH, `working <pid>` with the runner argv and wrapper checked, `done 3` after exit, and `cli_binary`/`state_file` overrides. 8 pass in ~9s. Not covered: the real installer and the real runner.

- Scripts run through `coder-utils` as an install step (no-op when the CLI is present) and a start step. Scripts, step logs, runner state, and runner log all live under `$HOME/.coder-modules/coder/<module>` for debugging.

## Type of Change

- [x] New module
- [ ] New template
- [ ] Bug fix
- [ ] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/agent-relay-claude-code`
**New version:** `v0.1.0`

## Testing & Validation

- [x] Tests pass (`bun test`, `terraform test`)
- [x] Changes tested locally (`readmevalidation`, `terraform validate`)

## Related Issues

None

---

Closes PLAT-462

### DIFF
--- a/registry/coder/modules/agent-relay-claude-code/README.md
+++ b/registry/coder/modules/agent-relay-claude-code/README.md
@@ -1,0 +1,87 @@
+---
+display_name: Agent Relay Claude Code
+description: Serves Claude Code self-hosted runner sessions in Coder workspaces that Agent Relay dispatches.
+icon: ../../../../.icons/claude.svg
+verified: true
+tags: [agent, claude, agent-relay]
+---
+
+# Agent Relay Claude Code
+
+Makes a Coder template a target for [Agent Relay](https://coder.com/docs/ai-coder/agent-relay)
+Claude Code pools. Agent Relay dispatches Claude Code sessions to workspaces
+built from the template; the module declares the parameters the relay stamps
+on each build and runs the Claude Code self-hosted runner.
+
+```tf
+module "claude_code_runner" {
+  source   = "registry.coder.com/coder/agent-relay-claude-code/coder"
+  version  = "0.1.0"
+  agent_id = coder_agent.main.id
+
+  # Downloads the Claude Code CLI at start when it is not in the image. Bake
+  # the CLI into the image and set this to false for faster workspaces.
+  install_cli = true
+}
+
+resource "coder_agent" "main" {
+  # ...
+  metadata {
+    key          = "agent_relay_status"
+    display_name = "Session"
+    script       = module.claude_code_runner.status_metadata_script
+    interval     = 10
+    timeout      = 5
+  }
+}
+```
+
+The `agent_relay_status` metadata block is required. It has to live on the
+`coder_agent`, which the module cannot declare; the relay reads it to decide
+when to reap the workspace.
+
+## Requirements
+
+- The `claude` CLI must be in the workspace. `install_cli` (default `true`)
+  downloads it at start only when it is not already on PATH; bake it into the
+  image for the fastest start. `cli_binary` overrides the path.
+- Builds must finish inside Agent Relay's 300s spawn budget: pre-pulled
+  images, no persistent volumes.
+
+## Parameters
+
+Agent Relay verifies this contract against the template's active version at
+startup and refuses to serve a pool that does not satisfy it. Every parameter
+renders disabled with a "Set by Agent Relay on dispatch" placeholder; the
+credential is masked.
+
+| parameter                                 | kind       | value                                                                        |
+| ----------------------------------------- | ---------- | ---------------------------------------------------------------------------- |
+| `agent_relay_session_id`                  | persistent | Anthropic session this workspace serves                                      |
+| `agent_relay_delivery_id`                 | persistent | work order that dispatched the build; rotates per attempt                    |
+| `agent_relay_pool`                        | persistent | Agent Relay pool that dispatched the build                                   |
+| `agent_relay_credential`                  | ephemeral  | single-use work order JWT (`SELF_HOSTED_RUNNER_POOL_SECRET`)                 |
+| `agent_relay_claude_code_lock_to_account` | ephemeral  | Anthropic account the runner locks to (`SELF_HOSTED_RUNNER_LOCK_TO_ACCOUNT`) |
+| `agent_relay_attempt`                     | ephemeral  | delivery attempt, echoed back when the relay nacks the session               |
+
+## Runner lifecycle
+
+The script starts `claude self-hosted-runner --capacity 1` detached and exits,
+so the agent reaches `ready` immediately. The runner is wrapped so every
+session runs with `--permission-mode bypassPermissions`; there is no terminal
+attached, so an approval prompt would hang it. When the runner exits, Agent
+Relay deletes the workspace; when it fails, the relay nacks the work order.
+
+`agent_relay_status` reports one of:
+
+| value             | meaning                                                                    |
+| ----------------- | -------------------------------------------------------------------------- |
+| `pending`         | no state recorded yet                                                      |
+| `idle`            | no credential: workspace was created manually                              |
+| `working`         | runner alive, no session picked up                                         |
+| `serving`         | runner alive, session picked up (best effort, see `serving_log_pattern`)   |
+| `orphaned`        | runner process gone without recording an exit                              |
+| `done <code>`     | runner exited with that status; `137` means it was killed                  |
+| `failed <reason>` | runner could not start, e.g. `runner-agent-missing` when the CLI is absent |
+
+Renaming the `agent_relay_status` key breaks reaping.
--- a/registry/coder/modules/agent-relay-claude-code/main.tf
+++ b/registry/coder/modules/agent-relay-claude-code/main.tf
@@ -1,0 +1,201 @@
+# Claude Code self-hosted runner module for Coder templates.
+#
+# A BYOC-compatible template includes this module and passes it the
+# workspace's coder_agent id. It declares every rich parameter
+# Agent Relay stamps on a build and runs `claude self-hosted-runner`
+# via a coder_script.
+#
+# Parameter contract (enforced by Agent Relay at startup via the dynamic
+# parameters evaluate endpoint):
+#
+#   - agent_relay_session_id, agent_relay_delivery_id, agent_relay_pool are
+#     persistent state: coderd only stores parameter values the
+#     template declares, and Agent Relay's dedupe and reconciliation query
+#     workspaces with `param:` search filters on them.
+#   - agent_relay_credential, agent_relay_claude_code_lock_to_account,
+#     and agent_relay_attempt are ephemeral runner inputs, reset
+#     between builds. An ephemeral value is still recorded on the build
+#     that set it, which is how Agent Relay reads the attempt back when
+#     it reaps a dead workspace.
+#
+# The runner's lifecycle is published through the agent_relay_status agent
+# metadata item, whose script this module renders. The script starts the
+# runner detached and exits so the agent reaches the ready lifecycle
+# state rather than sitting in starting for the whole session.
+
+terraform {
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = ">= 2.4.0"
+    }
+  }
+}
+
+variable "agent_id" {
+  type        = string
+  description = "ID of the coder_agent that should receive the runner env vars and run the runner script."
+}
+
+variable "cli_binary" {
+  type        = string
+  default     = "claude"
+  description = "Path to the Claude Code CLI binary in the workspace image. Override to test a beta build."
+}
+
+variable "install_cli" {
+  type        = bool
+  default     = true
+  description = "Install the Claude Code CLI (curl https://claude.ai/install.sh -fsSL | bash) when the workspace starts and the CLI is not already on PATH. Defaults to true so a template works against an image that has no CLI. A CLI already in the image is used as is and never upgraded, which is the recommended and fastest path: the download only runs when the binary is missing, and it spends part of the claim-to-ready window when it does."
+}
+
+variable "state_file" {
+  type        = string
+  default     = "/tmp/agent-relay/runner-state"
+  description = "Path the runner supervisor writes its lifecycle state to, read by the agent_relay_status agent metadata item."
+}
+
+variable "log_file" {
+  type        = string
+  default     = "/tmp/agent-relay/runner.log"
+  description = "Path the detached runner's output is written to."
+}
+
+variable "serving_log_pattern" {
+  type        = string
+  default     = "Picked up session"
+  description = "Runner log substring that means a session was claimed. Only distinguishes the agent_relay_status values working and serving; a stale pattern degrades to working and affects nothing else."
+}
+
+data "coder_parameter" "agent_relay_session_id" {
+  name         = "agent_relay_session_id"
+  display_name = "Agent Relay session"
+  description  = "Anthropic session this workspace serves. Agent Relay sets this when it dispatches the workspace; a human never fills it in. The relay uses it to recognize its own workspaces, dedupe redeliveries, and reconcile state after a restart."
+  type         = "string"
+  mutable      = true
+  default      = ""
+  order        = 1000
+  styling = jsonencode({
+    disabled    = true
+    placeholder = "Set by Agent Relay on dispatch"
+  })
+}
+
+data "coder_parameter" "agent_relay_delivery_id" {
+  name         = "agent_relay_delivery_id"
+  display_name = "Agent Relay delivery"
+  description  = "Work order that dispatched this build, identified by its JWT id. Agent Relay sets this when it dispatches the workspace; a human never fills it in. Tracing only: it rotates on every delivery attempt of the same session, so it identifies the delivery rather than the session."
+  type         = "string"
+  mutable      = true
+  default      = ""
+  order        = 1001
+  styling = jsonencode({
+    disabled    = true
+    placeholder = "Set by Agent Relay on dispatch"
+  })
+}
+
+data "coder_parameter" "agent_relay_pool" {
+  name         = "agent_relay_pool"
+  display_name = "Agent Relay pool"
+  description  = "Agent Relay runner pool that dispatched this build. Agent Relay sets this when it dispatches the workspace; a human never fills it in. One relay can serve several pools, each with its own Anthropic credential, organization, and template."
+  type         = "string"
+  mutable      = true
+  default      = ""
+  order        = 1002
+  styling = jsonencode({
+    disabled    = true
+    placeholder = "Set by Agent Relay on dispatch"
+  })
+}
+
+data "coder_parameter" "agent_relay_credential" {
+  name         = "agent_relay_credential"
+  display_name = "Agent Relay credential"
+  description  = "Single-use work order JWT the runner authenticates to Anthropic with. Agent Relay sets this when it dispatches the workspace; a human never fills it in. Ephemeral: it is valid for one session and is not reused on a later build."
+  type         = "string"
+  ephemeral    = true
+  mutable      = true
+  default      = ""
+  order        = 1003
+  styling = jsonencode({
+    disabled    = true
+    mask_input  = true
+    placeholder = "Set by Agent Relay on dispatch"
+  })
+}
+
+data "coder_parameter" "agent_relay_claude_code_lock_to_account" {
+  name         = "agent_relay_claude_code_lock_to_account"
+  display_name = "Claude Code account lock"
+  description  = "Anthropic account the runner is locked to, so a session can only ever be served for the account it was issued for. Agent Relay sets this from the work order's account_id claim; a human never fills it in."
+  type         = "string"
+  ephemeral    = true
+  mutable      = true
+  default      = ""
+  order        = 1004
+  styling = jsonencode({
+    disabled    = true
+    placeholder = "Set by Agent Relay on dispatch"
+  })
+}
+
+data "coder_parameter" "agent_relay_attempt" {
+  name         = "agent_relay_attempt"
+  display_name = "Agent Relay delivery attempt"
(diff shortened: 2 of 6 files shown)


registry/coder/modules/agent-relay-claude-code/README.md
```diff
@@ -0,0 +1,87 @@
+---
+display_name: Agent Relay Claude Code
```
matifali, line 2: @mattvollmer any suggestions on naming here from SEO perspective?

registry/coder/modules/agent-relay-claude-code/README.md
```diff
@@ -0,0 +1,87 @@
...
+renders disabled with a "Set by Agent Relay on dispatch" placeholder; the
+credential is masked.
+
+| parameter                                 | kind       | value                                                                        |
+| ----------------------------------------- | ---------- | ---------------------------------------------------------------------------- |
+| `agent_relay_session_id`                  | persistent | Anthropic session this workspace serves                                      |
+| `agent_relay_delivery_id`                 | persistent | work order that dispatched the build; rotates per attempt                    |
+| `agent_relay_pool`                        | persistent | Agent Relay pool that dispatched the build                                   |
+| `agent_relay_credential`                  | ephemeral  | single-use work order JWT (`SELF_HOSTED_RUNNER_POOL_SECRET`)                 |
+| `agent_relay_claude_code_lock_to_account` | ephemeral  | Anthropic account the runner locks to (`SELF_HOSTED_RUNNER_LOCK_TO_ACCOUNT`) |
```
matifali, line 64: it's good to know it's single use but how Template insights work we capture every unique value of a parameter. Not sure if it's the same for ephemeral parameters too. So I would suggest verifying.

registry/coder/modules/agent-relay-claude-code/main.tf
```diff
@@ -0,0 +1,201 @@
...
+
+variable "state_file" {
+  type        = string
+  default     = "/tmp/agent-relay/runner-state"
+  description = "Path the runner supervisor writes its lifecycle state to, read by the agent_relay_status agent metadata item."
+}
+
+variable "log_file" {
+  type        = string
+  default     = "/tmp/agent-relay/runner.log"
```
matifali, line 60: should we use a default thats likely to be persistent just in case if someone needs to debug?

registry/coder/modules/agent-relay-claude-code/README.md
Emyrk: It might. 

Passing secrets this way is not ideal

registry/coder/modules/agent-relay-claude-code/main.tf
Emyrk: We don't persist agent logs or anything else to the home directory as it stands.

All our logs go to /tmp by default. I hear this concern, and it is configurable :shrug:

35C4n0r: could there be a case where a template has both claude-code-relay and cursor-relay module ?

registry/coder/modules/agent-relay-claude-code/main.tf
```diff
@@ -0,0 +1,201 @@
...
+  display_name = "Claude Code runner"
+  icon         = "/emojis/1f916.png"
+  run_on_start = true
+  script = templatefile("${path.module}/run.sh.tftpl", {
+    cli_binary  = var.cli_binary
+    install_cli = var.install_cli
+    state_file  = var.state_file
+    log_file    = var.log_file
+  })
+}
```
35C4n0r, line 183: Let's use coder_utils and set the module_directory to `/tmp/agent-relay`. This will help us persist the run logs and the script for debugging.

registry/coder/modules/agent-relay-claude-code/run.sh.tftpl
```diff
@@ -0,0 +1,118 @@
...
+}
+
+if [ -z "$${SELF_HOSTED_RUNNER_POOL_SECRET:-}" ]; then
+	write_state "idle"
+	echo "No work order set. This workspace was created manually, not by Agent Relay."
+	echo "Idling. Delete this workspace manually or set the agent_relay_credential parameter."
+	exit 0
+fi
+
+%{ if install_cli ~}
```
35C4n0r, line 31: Write the logic unconditionally and execute conditionally.

35C4n0r (changes requested):

registry/coder/modules/agent-relay-claude-code/main.tf
Emyrk: TIL `coder_utils`. That is great. I will switch to this

registry/coder/modules/agent-relay-claude-code/main.tf
Emyrk: > could there be a case where a template has both claude-code-relay and cursor-relay module ?

It doesn't make sense to run both modules at the same time. But you could conditionally select 1 module or the other

registry/coder/modules/agent-relay-claude-code/main.tf
Emyrk: Done in 9b7d5ac. The script is now an install step (no-op when the CLI is present) and a start step, run through coder-utils.

One change to the ask: coder-utils only accepts `$HOME/.coder-modules/<ns>/<name>`, so the directory is `$HOME/.coder-modules/coder/agent-relay-claude-code` rather than `/tmp/agent-relay`. Runner state and log default there too, so scripts, step logs, runner log, and supervisor all sit in one tree.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+	exit 1
+fi
+
+# Write a wrapper that forces bypassPermissions so sessions never stall
+# on a tool-approval prompt. There is no terminal attached to the child
+# claude process, so an approval prompt would hang the runner.
+mkdir -p "$${HOME}/.claude"
+cat >"$${HOME}/.claude/wrapper.sh" <<'WRAPPER'
+#!/usr/bin/env bash
+exec "$CLAUDE_RUNNER_CLAUDE_BIN" "$@" --permission-mode bypassPermissions
```
matifali, line 72: **Blocker: this fails in any workspace whose agent runs as root.** The CLI refuses the flag as uid 0 — verified against `claude 2.1.161-byoc.14`:

```
# claude --print --permission-mode bypassPermissions 'hi'
--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons
```

Every session then dies immediately while the state file still says `working`, so the relay keeps feeding the workspace. `IS_SANDBOX=1` in the supervisor env clears the guard (children inherit it) — verified — or detect uid 0 here and `write_state "failed …"` so the relay nacks. The stub-`claude` tests can't catch this: `main.test.ts` runs as root and only asserts the wrapper's text.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/status.sh.tftpl
```diff
@@ -0,0 +1,56 @@
...
+
+case "$state" in
+working)
+	if [ -z "$${detail:-}" ] || ! kill -0 "$detail" 2>/dev/null; then
+		# The runner vanished without the supervisor recording an exit,
+		# e.g. an OOM kill. Distinct from done so Agent Relay can nack.
+		echo "orphaned"
+		exit 0
+	fi
+	if [ -f "$log_file" ] && grep -q "${serving_log_pattern}" "$log_file"; then
```
matifali: **Shell injection via `serving_log_pattern`** — it lands unquoted inside a command. Rendering with `serving_log_pattern = "x\"; touch /tmp/PWNED; \""` produced a status script that ran the `touch` and still printed `serving` (verified).

Per AGENTS.md's encoding table this is a free-form string: `base64encode` it in `main.tf`, decode into a variable here, and match with `grep -qF -- "$pattern"` so regex metacharacters are literal too.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+if [ -z "$${SELF_HOSTED_RUNNER_POOL_SECRET:-}" ]; then
+	write_state "idle"
+	echo "No work order set. This workspace was created manually, not by Agent Relay."
+	echo "Idling. Delete this workspace manually or set the agent_relay_credential parameter."
+	exit 0
+fi
+
+%{ if install_cli ~}
+# The install step may have placed the CLI under ~/.local/bin.
+export PATH="$HOME/.local/bin:$PATH"
```
matifali, line 41: `install_cli = false` drops this export (and the one at line 71) entirely — confirmed by diffing the two rendered scripts. So a CLI sitting at `~/.local/bin`, which is where the official installer puts it, isn't found and the module writes `failed runner-agent-missing` — that's the documented BYO-CLI path. Move the export outside the conditional; it's harmless when the directory doesn't exist.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+mkdir -p "$state_dir" "$(dirname "$log_file")"
+
+write_state() {
+	# Write via a temp file so a reader never sees a partial line.
+	printf '%s\n' "$1" >"$state_file.tmp"
+	mv "$state_file.tmp" "$state_file"
+}
+
+if [ -z "$${SELF_HOSTED_RUNNER_POOL_SECRET:-}" ]; then
+	write_state "idle"
```
matifali: This overwrites terminal state on a restart: starting from `done 3`, a re-run with no credential leaves the state file at `idle` (verified), which the README defines as "created manually, never had a session" — so a real failure reads as nothing-to-see-here. Only write `idle` when no state file exists.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+write_state() {
+	printf '%s\n' "\$1" >"$state_file.tmp"
+	mv "$state_file.tmp" "$state_file"
+}
+
+%{ if install_cli ~}
+export PATH="\$HOME/.local/bin:\$PATH"
+%{ endif ~}
+${cli_binary} self-hosted-runner \\
+	--capacity 1 \\
```
matifali, line 91: No idle bound. Per the runner's own help text, `--exit-if-unused-min` defaults to *never*, so a workspace that is dispatched but never receives a session stays `working` indefinitely. (`--drain-grace-sec` defaulting to 0 covers the after-a-session case, not this one.) #1112 passes an idle timeout — worth exposing one here too.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+
+write_state() {
+	printf '%s\n' "\$1" >"$state_file.tmp"
+	mv "$state_file.tmp" "$state_file"
+}
+
+%{ if install_cli ~}
+export PATH="\$HOME/.local/bin:\$PATH"
+%{ endif ~}
+${cli_binary} self-hosted-runner \\
```
matifali, line 90: `cli_binary` is interpolated unvalidated into the supervisor heredoc — same class as the `serving_log_pattern` issue, lower reachability. A `validation` block on the variable is probably enough.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
```diff
@@ -0,0 +1,103 @@
...
+	echo "The runner binary '${cli_binary}' is not available." >&2
+	echo "Bake the Claude Code CLI into the image, set install_cli = true, or set cli_binary. See the module README." >&2
+	exit 1
+fi
+
+# Write a wrapper that forces bypassPermissions so sessions never stall
+# on a tool-approval prompt. There is no terminal attached to the child
+# claude process, so an approval prompt would hang the runner.
+mkdir -p "$${HOME}/.claude"
+cat >"$${HOME}/.claude/wrapper.sh" <<'WRAPPER'
```
matifali: Now that everything else lives under `module_directory`, the wrapper should too (`scripts/wrapper.sh`). `~/.claude` is the runner's *host config dir*: it snapshots that tree into every child session's `CLAUDE_CONFIG_DIR`, and `wrapper.sh` isn't on its skip list, so a copy is seeded into each session.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/main.tf
```diff
@@ -0,0 +1,223 @@
...
+  source  = "registry.coder.com/coder/coder-utils/coder"
+  version = "0.0.1"
+
+  agent_id            = var.agent_id
+  module_directory    = local.module_directory
+  display_name_prefix = "Claude Code runner"
+  icon                = "/emojis/1f916.png"
+  install_script      = local.install_script
+  start_script        = local.start_script
+}
```
matifali, line 232: AGENTS.md asks modules wrapping `coder-utils` to re-export the pass-through so callers can serialize behind this pipeline:

```tf
output "scripts" {
}
```

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/main.tftest.hcl
```diff
@@ -0,0 +1,187 @@
...
+  }
+}
+
+run "install_cli_enabled_by_default" {
+  command = plan
+
+  # A CLI already in the image must short-circuit the download, so a
+  # prepared image spends none of the claim-to-ready window on it.
+  assert {
+    condition     = can(regex("if command -v claude >/dev/null 2>&1; then\n\techo \"Claude Code CLI already present", local.install_script))
```
matifali, line 157: Nit: this encodes rendered newlines and literal tabs, so reindenting the template breaks the test without changing behavior. Asserting the two distinctive substrings separately is enough.

---
_Generated by [Claude Code](https://claude.ai/code)_

matifali (commented):
Reviewed at `9b7d5ac`. Findings below were verified empirically rather than by reading: both `.tftpl` files rendered with real `terraform` (both `install_cli` values), the rendered scripts executed, and the runner contract checked against the shipped `claude 2.1.161-byoc.14` binary. `terraform test` passes 5/5.

Design looks sound — detached supervisor, state-file lifecycle, terminal `failed` state for a missing binary. Two I'd treat as blocking:

- **`bypassPermissions` + root** (`start.sh.tftpl:53`) — the CLI hard-refuses the flag as uid 0, so in a root workspace every session dies instantly while status still reads `working`.
- **Injection via `serving_log_pattern`** (`status.sh.tftpl:41`) — unquoted interpolation into a command; reproduced.

Then: `install_cli = false` loses the `~/.local/bin` PATH export (breaks the BYO-CLI path), `idle` overwrites terminal state on restart, and nothing bounds an unused runner's lifetime.

One gap not tied to a line: nothing executes `status.sh` — `main.test.ts` never runs it, so none of `pending / idle / working / serving / orphaned / done / failed` is covered. A test that walks the state file through each value is cheap and would have caught two of the above.

Happy to be wrong on any of these — say the word and I'll re-check.

---
_Generated by [Claude Code](https://claude.ai/code)_

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Fixed in 589173f. The wrapper exports `IS_SANDBOX=1` before exec, so only the session child sees it; the supervisor and runner env are unchanged. Chose this over failing on uid 0 since the runner workspace is a disposable container with nothing else in it, and refusing root would break the common agent-as-root image.

The test now runs the wrapper as root against a stub that enforces the CLI's guard; it fails without the export.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/README.md
Emyrk: I have a linear issue for this

matifali (approved):

registry/coder/modules/agent-relay-claude-code/status.sh.tftpl
Emyrk: Fixed in 30922fc: `base64encode` in `main.tf`, decoded into a variable in the status script, matched with `grep -qF -- \"$serving_log_pattern\"`. Added a tftest with your exact payload and a container test that runs the rendered status script against a log with and without the literal pattern, checking `working`/`serving` and that nothing executed.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Fixed in 790770f: both exports are unconditional now. Added a container test that sets `install_cli = false`, drops a stub CLI at `/root/.local/bin/claude`, and expects `working`; it reported `failed runner-agent-missing` before.

<sub>Coder Agents on behalf of @Emyrk.</sub>

matifali:
@Emyrk lets make our module always downlaod/update to the latest version too.
It may help address [RELAY-2](https://linear.app/codercom/issue/RELAY-2/handle-outdated-claude-code-binary-in-runner)

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Fixed in 8fb1d73: `idle` is written only when no state file exists. Added a container test that seeds `done 3`, runs the start step with no credential, and expects `done 3` to survive.

<sub>Coder Agents on behalf of @Emyrk.</sub>

Emyrk:
> @Emyrk lets make our module always downlaod/update to the latest version too. It may help address [RELAY-2](https://linear.app/codercom/issue/RELAY-2/handle-outdated-claude-code-binary-in-runner)

We can make this a followup, but in dogfood we will not turn on download.
It's faster to provision if we cache in the image. Downloading is not a recommended way

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Added in cd9ce7a: `exit_if_unused_min` (number, default 10, 0 disables) rendered as `--exit-if-unused-min N`. When it fires the supervisor records `done <code>` and the relay reaps as usual. Picked 10 to match the relay's dispatch deadline; the description tells template authors to keep it above that deadline.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Fixed in 30c7708: `cli_binary` now has a `validation` block allowing only `[A-Za-z0-9._/@+-]`, so a command name or path passes and anything with whitespace, quotes, or other metacharacters is refused at plan time. tftest `cli_binary_rejects_shell` uses the `serving_log_pattern` payload with `expect_failures`.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/start.sh.tftpl
Emyrk: Moved in d8563df. The wrapper is written next to `supervise.sh` under the module directory (`$state_dir/wrapper.sh`) and passed to `--exec-path` from there; nothing is written under `~/.claude` anymore. tftest guards against the old path coming back.

<sub>Coder Agents on behalf of @Emyrk.</sub>

registry/coder/modules/agent-relay-claude-code/main.tf
Emyrk: Added in 2fbed72, with a tftest that the output equals `module.coder_utils.scripts`.

<sub>Coder Agents on behalf of @Emyrk.</sub>

35C4n0r (approved):
LGTM from a code pov. @matifali if it runs as expected (I can't test it due to access issues), we should be good to merge.
One minor nit, I'd have put *.sh.tftpl files in a scripts folder, but this can be done later and is non-blocking.

PR #1133, opened 2026-09-21
refactor(templates): use code-server module instead of manual install

phorcys420:
Bundles the manual `code-server` installs across three templates to use the [`code-server`](https://registry.coder.com/modules/coder/code-server) module instead of hand-rolling the install in `coder_agent.startup_script`.

## Changes
- **`coder/kubernetes`** — replace the manual install + hand-written `coder_app` with the module (`folder = "/home/coder"`).
- **`coder/nomad-docker`** — drop the manual install; the template already declared the module, so code-server was installed and served twice on `:13337`.
- **`coder-labs/docker-build`** — replace the manual install (kept the `/etc/skel` first-start bootstrap) + hand-written `coder_app` with the module (`folder = "/home/${local.username}"`).

Templates only, so no module version bumps. `terraform fmt -check` and `terraform validate` pass for all three.

<details>
<summary>Context / decision log</summary>

From a registry-wide audit of template anti-patterns. Manually installing code-server in `startup_script` duplicates what the maintained `code-server` module already does (install, background process, `coder_app`, healthcheck) and drifts from every other template, which consume the module.

Behavior preserved:
- Port stays `13337` (module default).
- `folder` set to the path each template previously hard-coded in its `coder_app` URL.
- `display_name`, `share`, `subdomain`, and healthcheck now come from module defaults, equivalent to the removed `coder_app` blocks.

Out of scope: `coder-labs/docker-build` still sets `GIT_*` env directly (separate follow-up; #1132 only covered the `coder/` templates).
</details>

---
🤖 Generated by Coder Agents on behalf of @phorcys420.


matifali (approved):
The olny downside of this along with #1132 is a bit less frindly in air-gapped enviroenmtes. As they need to connect to our registry.

phorcys420:
I agree but at least we get consistency with the rest and we really need to figure out air-gapped registry access either way.

https://github.com/coder/registry/blob/12b66eae875eedfb2f283e321c50daec782a00c5/registry/coder/modules/code-server/run.sh#L111 should be made configureable too, it should not be forced to code-server.dev, similar to our other modules with configureable CDN issues.

phorcys420:
I agree but at least we get consistency with the rest and we really need to figure out air-gapped registry access either way. I am only unifying something that >80% of templates already do for the sake of good practices, but yeah this symptom of our bad air gap support sucks.

https://github.com/coder/registry/blob/12b66eae875eedfb2f283e321c50daec782a00c5/registry/coder/modules/code-server/run.sh#L111 should be made configureable too, it should not be forced to code-server.dev, similar to our other modules with configureable CDN issues.

matifali:
Terraform 1.15 has nice feature to provide module source as an input which can allow air-gapped users to provide their own internal module URL host as a variable input

phorcys420:
yes there is also some caching proxies for terraform

PR #1134, opened 2026-09-21
refactor(aws-nixos): split amazon-init out as a Nix-agnostic module

phorcys420:
Stacked on `phorcys/nixos-tests` (the `aws-nixos` template itself), so the diff here is only the split.

## Why

The template mixed two concerns that change for different reasons: **how a NixOS machine is configured from a flake**, and **how EC2 gets Coder onto an instance at all**. The second is the part that is not portable, and it exists entirely because the NixOS AMI runs `amazon-init.service` instead of cloud-init: one hook, no `runcmd`, no `write_files`, no once-vs-per-boot distinction, and nothing can be ordered `After` it without deadlocking the first boot.

## The contract

`modules/amazon-init/` is now Coder-on-an-amazon-init-AMI and nothing else. It takes **one opaque string**:

```tf
module "amazon_init" {
  source = "./modules/amazon-init"

  agent_token       = coder_agent.main.token
  agent_init_script = coder_agent.main.init_script
  boot_script       = local.boot_script
}
```

and on every boot it publishes the agent handoff, sets the hostname, registers a log source and writes the logging library, publishes `workspace.json`, writes any `files`, runs `boot_script` as a child process, and starts the agent — always, even when the boot script failed, because a workspace whose boot script broke is exactly the one you need a terminal on.

Nothing in the module mentions Nix, a flake, or a checkout. It reads `coder_workspace` and `coder_workspace_owner` itself, so the caller passes no identity. It owns `log.sh` outright and exposes it as `log_library` for callers that want `coder_log` in a `coder_script` of their own. The one image-specific thing it cannot know — how to obtain `curl` when the AMI has none — is `curl_resolve_command`, which the template fills in with `nix build nixpkgs#curl`.

Everything Nix now lives in `scripts/boot.sh.tftpl`, which is handed over as that string and is the only file that names the flake, the checkout or the state directories.

## Two bugs this shook out

- **base64 payloads broke the 16 KiB limit.** Embedding the libraries base64-encoded inflates them by a third and leaves gzip nothing to compress: user-data went from ~9 KiB to `18431` bytes and the precondition correctly refused the build. Interpolated as text it is ~10.5 KiB.
- **`CODER_LOG_READY` was per-process.** Once the boot script became a child process it logged into a void — the rebuild ran correctly and reported nothing at all. It is now exported and inherited, and the library documents that registration is per build, not per process.

## Verification

Built workspaces from the refactored template on a test deployment:

```
13:32:31 NixOS  Cloning https://github.com/coder/nixos-example-flake into /etc/nixos
13:32:32 NixOS  Applying /etc/nixos#workspace-x86_64 (c2a4363eb266)
13:32:45 NixOS  Building the system configuration...
13:33:57 NixOS  Switch complete.
13:33:57 code-server  Installing code-server!
13:34:00 NixOS  coder-agent is active.
```

Ordering is unchanged (rebuild, then agent, then startup scripts) and `workspace.json` is populated from the module's own data sources. `terraform validate`, `readmevalidation`, `shellcheck --severity=warning`, `validate_set_u_order.sh` and `bun run fmt` are clean.

Generated with [Xum](https://mux.coder.com/) using Claude.


phorcys420:
whoops didn't mean to pr

PR #1135, opened 2026-09-21
refactor(templates): use git-config module in community templates

phorcys420:
Follow-up to #1132, applying the same change to community-namespace templates: replace the hand-rolled `GIT_AUTHOR_*` / `GIT_COMMITTER_*` `coder_agent` env vars with the [`git-config`](https://registry.coder.com/modules/coder/git-config) module.

The module sets the same env vars, additionally runs `git config --global`, and exposes optional override parameters.

### Affected templates

- `sharkymark/docker-claude` — already had the `git-config` module; just dropped the now-redundant `env` block
- `joergklein/docker-texlive` — kept the other env vars (`HOME`/`USER`/`LANG`/`LC_ALL`), removed only the `GIT_*` ones
- `iamtaochen/ssh-linux`
- `matifali/omnigent-docker`
- `coder-labs/docker-build`
- `umair/proxmox-vm`
- `bpmct/railway`

`nboyers/cloud-dev` only *reads* `GIT_*` from env in its setup shell script and never sets them in Terraform, so it was left unchanged.

### Notes

- Pinned to `~> 1.0`; matched each template's existing module `count` / formatting conventions.
- The module defaults to `allow_username_change = true`, which adds a "Full Name for Git config" parameter to these templates.

### Verification

- `terraform fmt`: clean on all seven
- `terraform validate`: passes on all seven

---
🤖 Generated with Coder Agents


matifali (approved):
LGTM

PR #1138, opened 2026-09-21
feat(registry/coder/modules/aws-region): regions.json catalog, default_availability_zone output, optional parameter

phorcys420:
Refactors `aws-region` into a data-driven catalog (following the `aws-ec2-instance-type` pattern), adds region/AZ outputs, and lets templates use the module without rendering a parameter.

## What changed

- **`regions.json`**: the region catalog moved out of `main.tf`. Stores `value`, `name`, and a `country` code per region; loaded with `jsondecode(file(...))`.
- **Flags live in `main.tf`**: a `flags` map keyed by country code (e.g. `us`, `eu`, `jp`) maps to the emoji path, so icons are defined once instead of repeated per region in the JSON.
- **`default_availability_zone` output**: resolves the selected region to a concrete zone (e.g. `us-east-1a`), replacing the `"${region}a"` guess the AWS templates do inline.
- **`regions` output**: the full catalog keyed by region ID (`name`, `country`, `icon`, `default_availability_zone`).
- **`create_parameter` variable** (default `true`): set `false` to skip the `coder_parameter` and use the module for its outputs only; `value` then falls back to `var.default`.
- **Removed the Terraform test file** (per request; the module now ships no tests).
- **README**: documents the outputs, the outputs-only usage, and how to regenerate `regions.json` via the AWS CLI.
- Version bump `1.0.31` -> `1.1.0`.

No breaking changes to inputs. The output was renamed `availability_zone` -> `default_availability_zone` (it was not part of a prior release).

## Testing

- `terraform validate` clean
- `bun run fmt:ci` clean
- README validator (`cmd/readmevalidation`) passes
- Manual `terraform apply` sanity checks: parameter on/off, AZ resolution, empty-default fallback

<details>
<summary>Implementation notes &amp; decisions</summary>

- `regions.json` and the `flags` map were both derived from the original `main.tf` region map (icons decoded to their ISO country codes), so names/icons are byte-identical to the prior list. The `eu` grouping (all European regions share the EU flag) is preserved exactly.
- `default_availability_zone` is computed as `<region>a`. Every commercial AWS region exposes an `a` zone by name, so it is a safe default and matches what the AWS templates already do; the `default_` prefix signals it is a sensible default, not the only zone.
- `create_parameter = false` uses the standard `count`-toggle pattern; `value` falls back to `var.default`, and the `regions` catalog output is always available.
- Tests were removed at the author's request. CI's Terraform/TS test steps skip modules with no test files, so this stays green. Heads up: `CONTRIBUTING.md` still lists tests under "Every Module Must Have".
- Version bumped to `1.1.0` per CONTRIBUTING (feature = minor); latest release tag is `v1.0.31`.

</details>

---

> Generated by Coder Agents on behalf of @phorcys420.


phorcys420:
just need to test

registry/coder/modules/aws-region/README.md
```diff
@@ -14,31 +14,78 @@ the region closest to them.
 Customize the preselected parameter value:
 
 ```tf
-module "aws-region" {
+module "aws_region" {
```
matifali: why `_`?

matifali (approved):

PR #1141, opened 2026-09-22
feat(agent-relay-cursor): add credential_kind for shared service-account keys

Emyrk:
## Description

Agent Relay's per-pool `insecure_shared_token` (coder/agent-relay#70) stamps the team service-account key into the worker instead of a per-user sub-token, for CLI releases that refuse delegated tokens. The module handed every credential to the CLI as `--auth-token`, which the CLI only accepts for worker tokens; with the shared key the worker died on start:

```
Error: Failed to validate worker account settings: If you are logged in, try logging out and back in.
```

Verified on dogfood: the same key as `CURSOR_API_KEY` is classified `team_api_key`, authenticates, and registers to the pool.

- New persistent parameter `agent_relay_cursor_credential_kind` (`worker_token` default, `api_key`), exported as `AGENT_RELAY_CURSOR_CREDENTIAL_KIND`. Agent Relay stamps it from the pool's `insecure_shared_token` (relay PR to follow), so no template variable has to be kept in step with pool config.
- Supervisor branches at run time: `api_key` exports `CURSOR_API_KEY="$AGENT_RELAY_CURSOR_TOKEN"` and omits `--auth-token`; `worker_token` is unchanged. Both read the value from the environment; nothing lands in the file.
- Defaults to `worker_token`, so a relay that does not stamp the parameter yet behaves exactly as before.
- README documents the pairing and the exposure it implies.
- tftest covers the parameter, its env export, and both rendered paths. Container test drives the supervisor with `AGENT_RELAY_CURSOR_CREDENTIAL_KIND=api_key`: stub worker sees `CURSOR_API_KEY`, no `--auth-token` in argv, literal key absent from `supervise.sh`.

## Type of Change

- [ ] New module
- [ ] New template
- [ ] Bug fix
- [x] Feature/enhancement
- [ ] Documentation
- [ ] Other

## Module Information

**Path:** `registry/coder/modules/agent-relay-cursor`
**New version:** `v0.2.0`

## Testing & Validation

- [x] Tests pass (`terraform test` 10 runs, `bun test` 18 tests)
- [x] Changes tested locally

## Related Issues

coder/agent-relay#70

---


matifali:
Looks good to me. Does it mean that Cursor provided two methods to authenticate? Can we separate them out in the UI and suggest users to use which one based on their scenario?

matifali (approved):

Emyrk:
> Looks good to me. Does it mean that Cursor provided two methods to authenticate? Can we separate them out in the UI and suggest users to use which one based on their scenario?

Yes there is 2 methods. The user token auth is not working anymore for some reason.
Ideally we don't expose this to the user. It's configured by the agent-relay

matifali:
By User I mean the Template teams.

